PR #6, opened 2019-03-14
feat(oauth): Get Oauth scaffolding in place & add email auth

ibolton336:


andyvesel:
@ibolton336 
Since I was trying to run it locally, I'm getting an empty screen and error
```
Uncaught SyntaxError: Unexpected token i in JSON at position 0
    at JSON.parse (<anonymous>)
    at eval (webpack-internal:///./src/app/auth/duck/reducers.ts:25)
    at Module../src/app/auth/duck/reducers.ts (app.bundle.js:23182)
    at __webpack_require__ (app.bundle.js:1427)
    at fn (app.bundle.js:801)
    at eval (webpack-internal:///./src/app/auth/duck/index.ts:2)
    at Module../src/app/auth/duck/index.ts (app.bundle.js:23158)
    at __webpack_require__ (app.bundle.js:1427)
    at fn (app.bundle.js:801)
    at eval (webpack-internal:///./src/reducers.ts:4)
```
Does it work fine for you?

src/app/home/HomeComponent.tsx
```diff
@@ -1,21 +1,211 @@
...
+} from "@patternfly/react-core";
+import { BellIcon, CogIcon } from "@patternfly/react-icons";
+import { Dimmer, Loader, Image, Segment } from "semantic-ui-react";
+
 import "./HomeComponent.css";
-class HomeComponent extends React.Component<any, any> {
+
+export default class HomeComponent extends React.Component<any, any> {
+  componentDidMount() {
+    console.log("this.state", this.state, this.props);
```
alexander-demichev: Do we want to have `console.log`  here?

src/app/auth/PrivateRoute/index.tsx
```diff
@@ -1,22 +1,21 @@
...
-      )
-    }
-  />
-);
+const PrivateRoute = ({ component: Component, ...rest }) => {
+  return (
+    <Route
+      {...rest}
+      render={props =>
+        localStorage.getItem("currentUser") ? (
```
bond95: Why you use `localStorage`, instead of redux? This is anti-pattern of React-Redux, because everything that you can move to redux, must be moved to redux. Otherwise it could break immutable logic of application.

src/app/auth/duck/operations.ts
```diff
@@ -13,36 +15,69 @@ const failedTokenRequest = Creators.failedToken;
...
+        //   Authorization: "Token" + hardCodedToken
+        // },
+        // mode: "cors"
+      })
+        .then(handleResponse)
+        // .then(response => response.json())
+        .then(res => {
+          localStorage.setItem("currentUser", JSON.stringify(res.user));
+          dispatch(loginSuccess(res.user.email));
+          history.push("/");
```
bond95: Why you use `history.push` if you can do this through redux and `connected-react-router`?

src/app/auth/duck/operations.ts
```diff
@@ -13,36 +15,69 @@ const failedTokenRequest = Creators.failedToken;
...
+    dispatch(logout());
+    localStorage.removeItem("currentUser");
+  };
+}
+
+const setOAuthTokenRequest = res => {
+  return dispatch => {
+    localStorage.setItem("currentUser", JSON.stringify(res));
+    dispatch(loginSuccess(res.name));
+    history.push("/");
```
bond95: Same question as above

src/reducers.ts
```diff
@@ -2,11 +2,9 @@ import { combineReducers } from "redux";
 import homeReducer from "./app/home/duck";
 import authReducer from "./app/auth/duck";
 import commonReducer from "./app/common/duck";
-import { connectRouter } from "connected-react-router";
-import { history } from "./app/common/duck/utils";
+// import { history } from "./app/common/duck/utils";
```
bond95: I think this comment must be removed.

src/app/auth/PrivateRoute/index.tsx
eriknelson: The current user state needs to be persisted, which is why it's used here. The exact implementation of this may change as we implement the real oauth flows, but this is fine for now.

bond95: And? In my first PR, I fixed this, and everything work well through Redux, why not to keep this?
https://github.com/bond95/mig-ui/blob/5971ea8f81c094e7b6259e7d3588d77a56ba545b/src/components/PrivateRoute/index.tsx#L4-L16

ibolton336:
> The main question is why you removed `connected-react-router`, because it mean that instead of redux immutable logic you use some `localStorage`, `history.listen` and `history.push`, which breaks main aim of redux, that is to saving global state of application, and when you some state part save not to redux it make debug more complicated, and in this case I don't understand why do you use redux in this application at all if that other state part can be rewritten without redux.

https://redux.js.org/advanced/usage-with-react-router

bond95:
@ibolton336 Why to write own util, if there is already fully-worked util, that already was used in this project? https://github.com/supasate/connected-react-router

src/app/AppComponent.tsx
```diff
@@ -1,27 +1,38 @@
...
-      )}
-    </Switch>
-  </div>
-);
+class AppComponent extends React.Component<any, any> {
+  constructor(props) {
+    super(props);
+
+    const { clearAlerts } = this.props;
+    history.listen((location, action) => {
```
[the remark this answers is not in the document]
ibolton336: I havent seen any edge cases where we could need redux router. What would be the benefit of adding it over just using react-router for now?

package.json
```diff
@@ -33,7 +38,6 @@
   "dependencies": {
     "@patternfly/react-core": "^2.1.4",
     "@types/react": "^16.8.7",
-    "connected-react-router": "^6.3.2",
```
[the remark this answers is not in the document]
ibolton336: Is there any reason to keep it? I didnt see any cases yet where we could need it?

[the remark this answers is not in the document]
bond95: Yes, for example `push`, or history listening.

src/app/AppComponent.tsx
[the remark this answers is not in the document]
bond95: Because adding redux router means, that all state of application will be in redux, which means that application can react on events inside of React-Redux structure, so in this case better question is, why to use Redux at all, if you won't use it's benefits?

src/app/auth/PrivateRoute/index.tsx
ibolton336: authentication did not persist through a page refresh with your implementation

ibolton336:
> @ibolton336
> Since I was trying to run it locally, I'm getting an empty screen and error
> 
> ```
> Uncaught SyntaxError: Unexpected token i in JSON at position 0
>     at JSON.parse (<anonymous>)
>     at eval (webpack-internal:///./src/app/auth/duck/reducers.ts:25)
>     at Module../src/app/auth/duck/reducers.ts (app.bundle.js:23182)
>     at __webpack_require__ (app.bundle.js:1427)
>     at fn (app.bundle.js:801)
>     at eval (webpack-internal:///./src/app/auth/duck/index.ts:2)
>     at Module../src/app/auth/duck/index.ts (app.bundle.js:23158)
>     at __webpack_require__ (app.bundle.js:1427)
>     at fn (app.bundle.js:801)
>     at eval (webpack-internal:///./src/reducers.ts:4)
> ```
> Does it work fine for you?

try this again with the json server. instructions added to the readme

src/app/auth/PrivateRoute/index.tsx
bond95: This is strange because this work, could you, please, in detail describe why it's not working for you?

PR #7, opened 2019-03-14
Added component types

bond95:
Added component props types and state types, and small refactor.


eriknelson:
@bond95 needs a rebase

bond95:
@eriknelson done, please, review.

eriknelson:
@bond95 looks like it's failing some linter checks: https://travis-ci.org/fusor/mig-ui/jobs/508503687

`yarn lint` will lint it, there's also a fix feature that will auto apply a fix for things like semicolons with tslint, I need to add a target to run this.

bond95:
> I see the containers with redux connect were deleted and the connect was folded into the related components. @bond95 can you elaborate on the motivations for this? Is that the pattern we want to follow going forwards?

The main point of this pattern is not to create files just to have files. If component has less then 10 lines and used only in one component, then it doesn't make any sense to create separate file for that. The same with `connect`, `connect` is too small to have separate file.
But it could work in cases when component is used separately from it 'connected' version, this is not that case, so yeah, I prefer this pattern.

> master has auth/PrivateRoute/index.tsx and this is just introducing a auth/PrivateRoute.tsx. We need to decide which one makes more sense and delete the other.

This pattern is similar. If component has only one file, than creating separate folder for it with only one file is useless.

 > You're calling all the prop types IProps and IState. My default feeling here is they should be more specific names like ILoginProps, where the former is implicitly understood to describe the component in the file. Personally, I prefer fully explicit names, but interested in Ian's feedback.

I prefer names like `IProps` and etc. Because component file has only one main component in it, which means that names like `IProps` by default refers to that component. (In case if there will be some other component that satisfies condition from the first paragraph, that component must have types with more specific names)

src/app/auth/LoginComponent.tsx
```diff
@@ -170,4 +178,14 @@ class LoginComponent extends React.Component<any, any> {
     );
   }
 }
-export default LoginComponent;
+
```
ibolton336: I dont really see a reason to move this here. Since we want to limit the amount of components that know about the redux store, this is only happening in 3 or 4 components anyway. Here is an article about the benefits of separating this logic out into its own container. It is a common pattern. https://www.fullstackreact.com/p/using-presentational-and-container-components-with-redux/

bond95: Yes, but this article describe benefits in case generic components, `LogicComponent` is not generic. So as wrote previously: But it could work in cases when component is used separately from it 'connected' version, this is not that case, so yeah, I prefer this pattern.

ibolton336: ok thats no problem. we can combine for now & seperate out later if we need to.

PR #8, opened 2019-03-15
Add AboutModal

andyvesel:
- Added basic mockup for `AboutModal`
- Added `images.d.ts`, which handles images import for TypeScript


src/app/modals/BasicAboutModal.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import { AboutModal, Button, TextContent, TextList, TextListItem } from '@patternfly/react-core';
+import brandImg from './pf_mini_logo_white.svg';
+import logoImg from './pf_logo.svg';
+
+class BasicAboutModal extends React.Component <{}, {isModalOpen: boolean}>{
```
bond95: Please, define types separately. Like here for example https://github.com/fusor/mig-ui/blob/736d36663cd99bcc96b162050b3dbd9da3d8ec58/src/app/AppComponent.tsx#L9

src/app/modals/BasicAboutModal.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
```
bond95: Please, for consistency use double quotes "

eriknelson: This is a style choice we should discuss as a team and enforce with pre-merge linting (I'm fine with either).

andyvesel: I've intended to replace all unneeded double quotes to single in further PRs, so maybe we want to keep the single quote convention?

eriknelson: They are functionally the same, and single quotes are more common. I've always preferred them when writing JS, so that would take my vote. I would suggest ignoring the issue for this PR, and following up with another PR that converts the project to whatever it is that we decide, and also introducing linting that enforces the rule and is integrated with CI.

eriknelson:
Is this intended to be used as a generic modal component? Feels strange to have that be an "AboutModal" if that's the case.

andyvesel:
@eriknelson no it supposed to be just a modal component for 'About' info. For other modals I intended to use other, that are based on an imported generic modal

src/app/modals/BasicAboutModal.tsx
andyvesel: Added

bond95:
@andyvesel I don't see any usage of About modal, could you put it somewhere?

eriknelson:
Why are we writing an About modal? AFAIR, it's not part of the mocks, so I'm unclear how this fits in.

andyvesel:
@eriknelson I'm reworking all this to make just a basic modal, which supposed to add a cluster (or so far just mocked data). Agreed, we don't need `About` modal at this moment, according to our mockups.

eriknelson:
Thanks @andyvesel, so the goal with this PR is to add a generic modal component that we can reuse for all our modal requirements?

andyvesel:
I think, the goal here is to add `AddCluster` modal, with two content fields and two buttons. Am I thinking correctly, that all generic components we can just import from `@patternfly/react-core`?
@bond95 Your opinion?

eriknelson:
Okay, just want to understand what this PR is looking to accomplish. If PF
has a generic modal, then it sounds like this is introducing the
AddClusterModal to the project.

andyvesel:
@ibolton336 I think, this PR is still yet to be fixed according to comments above. I believe it's too early for merge

PR #13, opened 2019-03-19
Update modal window

andyvesel:
Updated modal window

- Added buttons
- Added text fields


andyvesel:
@bond95 should that be like `src/app/cluster/modals/AddClusterModal` thing?

bond95:
> @bond95 should that be like src/app/cluster/modals/AddClusterModal thing? Or modals directory is unneeded?

Modal directory is needed only in case if there will be more than one modals for clusters.

andyvesel:
@bond95 fixed

src/app/cluster/AddClusterModal.tsx
```diff
@@ -0,0 +1,35 @@
+import {
+  Modal,
+  Button,
+  TextInput,
+  TextContent,
+  TextList,
+  TextListItem,
+  TextArea,
+} from '@patternfly/react-core';
+import React from 'react';
```
bond95: Please, move this import to first line, to keep consistency.

src/app/cluster/AddClusterModal.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const AddClusterModal = ({ onHandleModalToggle, isModalOpen, ...props }) => (
```
bond95: Please, add type check.

src/app/cluster/AddClusterModal.tsx
eriknelson: We need to decide if we care about import ordering and if so, codify that in the linter.

bond95: I don't so much care about import ordering, but React import first is something like `unspoken rule`. If you look at code in other React projects or at least at tutorial React import in most cases is in first place.

andyvesel: Ok, fixed

src/app/cluster/AddClusterModal.tsx
andyvesel: Added

eriknelson:
@andyvesel needs rebase

andyvesel:
@eriknelson done

PR #19, opened 2019-03-22
Add styles & features to datalist view

ibolton336:


config/config.dev.json.example
```diff
@@ -1,4 +0,0 @@
-{
```
eriknelson: This file shouldn't be deleted

Issue #21, opened 2019-03-22
What exactly should be capitalized?


eriknelson:
I'm noticing by default `reduxsauce` produces `Creators` and `Types`, which we re-export. Similarly, all the Components are capitalized, but other types are not. It's feeling a bit random, so I'd like us to consider what convention we should establish for what should and should not be capitalized?

eriknelson:
Closing this as I don't think it's a pressing issue. Please reopen if someone wants to tackle this.

PR #22, opened 2019-03-22
Add OAuth and MockLogin flows

eriknelson:
This PR cleans up and refactors auth to use two different LoginComponents depending on if you are running in a local development configuration. In the local case, the `MockLoginComponent` will be substituted in during the webpack build. It simply logs in the user with more mock data so the developer can focus on the UI. In the remote & production cases, a full code grant oauth flow is implemented. This needs to be fleshed out and completed, but this lays the split groundwork.

I'm also noticing what seems like random capitalization choices, so I filed an issue (#21) to address it and we can submit a follow up PR to apply consistency.


ibolton336:
LGTM 👍

PR #28, opened 2019-03-25
Remove unneeded elements

andyvesel:
Removed unneeded dropdown menus


eriknelson:
Going to close this since we want to keep those there until it's determined they're not required.

PR #30, opened 2019-03-25
Add & remove cluster functionality

ibolton336:


andyvesel:
One general suggestion for future contributions is to split all the pushed stuff to relevant commits. That will be much easier to review. Also, please, add a description to this PR.

PR #37, opened 2019-03-30
[DO NOT MERGE] Introduce client and mock support

eriknelson:
Wanted to post this since the core of the client is there even though I'm still working through the touchpoints with the more front-end code.

`jsonServer` does not integrate well with kube objects, so I added a `KubeStore` which functions as an in-memory mock server that plugs into the client's interface by default. It'd remove the dependency on a mock server at all, although we probably want a way to serialize/deserialize its state into a db file, and a way to easily pre-load it from a persisted file.

Might make sense to chop this up into finer PRs as well, figure we can discuss Monday.

Dumping some more thoughts:

While writing this, it became clear there's a duplication of code in the operations; if nothing changes, we'd going to continue to write them with crud ops that look the same, barring a few exceptions (adding a cluster is a composite action that requires a cluster registry object + a secret). We should discuss a longer term solution if this continues to be the case, but we also need to ship something that works right now.

We shouldn't be adding distinct `duck/types.ts` files with type consts. They're produced by `reduxsauce` (if we want to continue to use it and its `Creators` and `Types`).

There are a couple renames that are significant:
* `MigrationCluster`: this is a `Cluster` now. We are trying to share the kubernetes federation object that already captures our semantic requirements. It's in a different API group than the mig objects.
* We've renamed most of our objects to be prefixed with a branded `Mig`. The reason is that `Migration` is fairly generic and results in objects named like `MigrationMigration` which is obviously not ideal. This impacted `migrationStorage` and should affect everything going forwards. If we're talking about one of our data models, it's prefixed with `migX`.


src/app/cluster/duck/operations.ts
```diff
@@ -1,73 +1,61 @@
...
-      },
-      error => {
-        dispatch(AlertCreators.alertError('Failed to add cluster'));
-      },
-    );
+const addCluster = cluster => {
+  return (dispatch, getState) => {
+    try {
+      const { migMeta } = getState();
+      const client: IClusterClient = ClientFactory.hostCluster(getState());
```
ibolton336: 👍

eriknelson:
Closing because we split this into #40 and #42 for easier review.

PR #42, opened 2019-04-01
Add client integration

eriknelson:
Also:
* Object renames to reflect those we'll be working with
* Removing some unecessary things like unused operations

Part 2 of splitting up #37 into specific PRs. This includes the integration with the new client.


src/app/cluster/duck/operations.ts
```diff
@@ -1,73 +1,61 @@
-import axios from 'axios';
 import { Creators as AlertCreators } from '../../common/duck/actions';
 import { Creators } from './actions';
-import { JSON_SERVER_URL } from '../../../config';
+import { ClientFactory } from '../../../client/client_factory';
```
ibolton336: nicely done!~

PR #46, opened 2019-04-01
Add wizard for adding mig plans

ibolton336:


src/app/plan/duck/actions.ts
```diff
@@ -1,5 +1,10 @@
 import { createActions } from 'reduxsauce';
 
-const { Creators, Types } = createActions({});
+const { Creators, Types } = createActions({
+  migrationPlanFetchSuccess: ['migrationPlanList'],
```
eriknelson: Think this should be `migPlan`, we should generally use `migX` when talking about the actual data objects that are getting posted, it's what we settled on as our prefix for our kube objects. The reason was, `MigrationMigration` is pretty weird, and we've been somewhat branding ourselves as "Mig", so we went with that. Ex: `MigPlan, MigCluster, MigStage, MigMigration`.

src/app/plan/duck/operations.ts
```diff
@@ -0,0 +1,70 @@
...
+import { IClusterClient } from '../../../client/client';
+import { MigResource, MigResourceKind } from '../../../client/resources';
+
+const migPlanFetchSuccess = Creators.migPlanFetchSuccess;
+const addPlanSuccess = Creators.addPlanSuccess;
+const addPlanFailure = Creators.addPlanFailure;
+const removePlanSuccess = Creators.removePlanSuccess;
+const removePlanFailure = Creators.removePlanFailure;
+
+const addPlan = migPlan => {
```
eriknelson: Nice :+1: 

Still irks me that these have started to really all look the same. There will be some divergence in some places, like `AddCluster`, since the act of adding a cluster actually requires the creation of 3 distinct objects.

There's probably a nice way for us to clean this up, but it's a great example of something that is *not* on the critical path right now, so we should file an issue and circle back at another point in time. I'll file it.

src/app/plan/duck/types.ts
```diff
@@ -0,0 +1,13 @@
+const FETCH_DATA = 'FETCH_DATA';
```
eriknelson: This file shouldn't exist; should use the Types generated from the reduxsauce actions file.

src/models.ts
```diff
@@ -6,7 +6,7 @@ interface IMigrationClusterMetadata {
     'migrations.openshift.io/migration-group': string;
   };
   name: string;
-  namespace: string;
+  namespaces: any[];
```
eriknelson: A namespaced object can only ever exist in a single namespace.

src/app/plan/components/GeneralForm.tsx
```diff
@@ -0,0 +1,48 @@
...
+  ...rest
+}) => (
+  <React.Fragment>
+    <Box>
+      <TextContent>
+        <TextList component="dl">
+          <TextListItem component="dt">Plan Name</TextListItem>
+          <input
+            onChange={handleChange}
+            onInput={() => setFieldTouched('planName', true, true)}
```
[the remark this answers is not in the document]
ibolton336: Yeah i literally just picked up the name right from the step title in the wireframes. Admittedly, It is a poor name.

src/models.ts
eriknelson: Didn't realize this was mocking out the list of namespaces from a `MigCluster`. I didn't make the KubeStore cluster aware, so it dumps everything into a single tree. Let's just use this for now and I'll fix the KubeStore to be able to handle this.

ibolton336:
> Just adding here so I don't forget: we need a bunch of functions that will take user input like form values and convert them into actual kube objects we can create. 

can we just use the form state from formik for this? there is a values object that accumulates all of the form values from the wizard form and allows you to manipulate them as needed. The wrapping "Wizard.tsx" component holds this formik state & any child components (i.e the forms for each page in the wizard) receive the formik event handlers, error object, & values object as props. From here we can also wrap up anything and send it to an operation with a redux action. 


> I think you've started to go down that path with models, but it's the function that acts as the glue between "I've got these form values the user gave me, but it's a handful of strings" => "I have a valid kube object that I can give to the client to post to the api server". Let's chat about this tomorrow.

eriknelson:
> can we just use the form state from formik for this? there is a values object that accumulates all of the form values from the wizard form and allows you to manipulate them as needed. The wrapping "Wizard.tsx" component holds this formik state & any child components (i.e the forms for each page in the wizard) receive the formik event handlers, error object, & values object as props. From here we can also wrap up anything and send it to an operation with a redux action.

Yeah that should work, just thinking we want to drop these functions in their own module or something to abstract it out of the components. These things keep changing so it makes it easier if they're all in one area and the presentation layer doesn't care at all about what they hand back, since it immediately hands it off to the operation to deal with.

PR #60, opened 2019-04-04
[DO NOT MERGE] Refactor: rewrote Promises to async/await

bond95:
Hey, I replaced Promises in `operations` to async/await. I think it is better to start using this feature earlier, because otherwise soon it there will be a lot of Promises and code turns into unreadable "callback hell". We had same situation in oVirt VM Portal, and we rewrote all Promises to `sagas` (it is similar solution as async/await, but based on generator `yield`). So it is better to start to use this feature now, to not get back here when it becomes an issue.

Also one of advantages is possibility to handle (catch) not only async errors, but sync too.

cc @eriknelson @ibolton336 @andyvesel

### DIFF
--- a/src/app/plan/duck/operations.ts
+++ b/src/app/plan/duck/operations.ts
@@ -16,7 +16,7 @@ const removePlanFailure = Creators.removePlanFailure;
 const sourceClusterNamespacesFetchSuccess = Creators.sourceClusterNamespacesFetchSuccess;
 
 const addPlan = migPlan => {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     try {
       const { migMeta } = getState();
       const client: IClusterClient = ClientFactory.hostCluster(getState());
@@ -24,10 +24,8 @@ const addPlan = migPlan => {
         MigResourceKind.MigPlan,
         migMeta.namespace,
       );
-      client
-        .create(resource, migPlan)
-        .then(res => dispatch(addPlanSuccess(res.data)))
-        .catch(err => AlertCreators.alertError('Failed to add plan'));
+      const res = await client.create(resource, migPlan);
+      dispatch(addPlanSuccess(res.data));
     } catch (err) {
       dispatch(AlertCreators.alertError(err));
     }
@@ -50,7 +48,7 @@ const removePlan = id => {
 };
 
 const fetchPlan = () => {
-  return (dispatch, getState) => {
+  return async (dispatch, getState) => {
     try {
       const { migMeta } = getState();
       const client: IClusterClient = ClientFactory.hostCluster(getState());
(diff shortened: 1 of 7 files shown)


src/app/cluster/duck/reducers.ts
```diff
@@ -58,7 +58,7 @@ export const INITIAL_STATE = {
 };
 
 export const clusterFetchSuccess = (state = INITIAL_STATE, action) => {
-  return { ...state, clusterList: action.clusterList };
+  return { ...state, clusterList: action.clusterList || [] };
```
litleleprikon: Probably should be in separate commit because this is functional change

src/app/plan/duck/operations.ts
```diff
@@ -11,18 +11,17 @@ const removePlanSuccess = Creators.removePlanSuccess;
...
       const resource = new MigResource(
         MigResourceKind.MigPlan,
         migMeta.namespace,
       );
-      client
-        .create(resource, migPlan)
-        .then(res => dispatch(addPlanSuccess(res.data)))
-        .catch(err => AlertCreators.alertError('Failed to add plan'));
+      const res = await client
+        .create(resource, migPlan);
```
litleleprikon: Minor: why do you call this method on new line?

src/app/storage/duck/reducers.ts
```diff
@@ -7,7 +7,7 @@ export const INITIAL_STATE = {
 };
 
 export const migStorageFetchSuccess = (state = INITIAL_STATE, action) => {
-  return { ...state, migStorageList: action.migStorageList };
+  return { ...state, migStorageList: action.migStorageList || [] };
```
litleleprikon: Same issue as in https://github.com/fusor/mig-ui/pull/60/files#r272197725

eriknelson:
@bond95 thanks for this, I'm being convinced we should use this going forwards, particularly when loading nested resource refs.

Can you run `yarn lint:fix` and push to pass CI?

eriknelson:
Was going to merge this and fix the lint with a follow up PR, but it's actually crashing the UI. As this is written, the AlertCard isn't capable of handling the error it's getting, I suspect:

```
api.ernelson.mg.dog8code.com:6443/apis/migration.openshift.io/v1alpha1/namespaces/mig/migstorage:1 GET https://api.ernelson.mg.dog8code.com:6443/apis/migration.openshift.io/v1alpha1/namespaces/mig/migstorage 404
react-dom.development.js?4646:55 Uncaught Invariant Violation: Objects are not valid as a React child (found: Error: Request failed with status code 404). If you meant to render a collection of children, use an array instead.
    in div (created by Context.Consumer)
    in AlertCard (created by AppComponent)
    in div (created by Context.Consumer)
    in Styled(div) (created by AppComponent)
    in div (created by Context.Consumer)
    in Styled(div) (created by AppComponent)
    in AppComponent (created by Context.Consumer)
```

I don't have time to investigate this currently, but I should be able to some time next week. Until that can get fixed, we shouldn't merge this.

eriknelson:
Blocked on #61

PR #74, opened 2019-04-10
Check connection button

andyvesel:
Added check connection button to Cluster modal, added redux


eriknelson:
We have the expected feedback described in the mockups.

eriknelson:
I'm merging this to master as is so we can pick it up in the demo branch. Fixes will be applied before demo is brought back to master.

PR #124, opened 2019-04-25
remove CardBody from detail view item

mcoker:


eriknelson:
Looks much better, thank you @mcoker!

PR #125, opened 2019-04-25
update add cluster, add repository modals

mcoker:
Removes border under title of modals, left aligns modal actions, reverses button order.


eriknelson:
Pushed the modal header background color change from this PR, but going to close because @ibolton336 is refactoring the modals and has already done the button alignment.

Issue #97, opened 2019-04-16
UX issues from 4/15 review


vconzola:
**_This_** means it should be fixed for Summit demo , if possible.

1. **_Need a title for the tool._**
2. Remove menu icon from the upper left because we don't have a left navigation.
3. Need the first entry empty state to display by default.
4. **_"Storage" section of the listview should be called "Replication repositories"_**
5. **_"Plans" section of the listview should be called "Migration plans"_**
6. **_Plans twistie needs an empty state if no plans have been created._**
7. Create Plan (link and button in the empty state) should be disabled unless two clusters and a repository have been created.
8. Only one section of the accordion should be able to be opened at one time.
9. Remove kebabs from the cards at the top.
10. **_When a cluster/repository/migration plan is created listview should open to that section when the modal/wizard closes._**
11. Wizard step 1 - Validate that the Name field is not empty before enabling Next button.
12. Wizard step 2 - Validate that a cluster and at least one project have been selected before enabling Next button.
13. Wizard steps 2, 3, 4 - Left align table column headers.
14. **_Wizard step 3 - All selectable PVs should default to Type = Copy._**

vconzola:
On the latest (4/24) code it looks like the following issues have been resolved:

1. Need a title for the tool.
2. Remove menu icon from the upper left because we don't have a left navigation.

4. "Storage" section of the listview should be called "Replication repositories"
5. "Plans" section of the listview should be called "Migration plans"
6. Plans twistie needs an empty state if no plans have been created.

9. Remove kebabs from the cards at the top.
10. *** (Only seems to have been fixed for Plans).
11. Wizard step 1 - Validate that the Name field is not empty before enabling Next button.

14. Wizard step 3 - All selectable PVs should default to Type = Copy.

eriknelson:
Thanks for the update Vince, that leaves:
3) Need the first entry empty state to display by default.
7) Create Plan (link and button in the empty state) should be disabled unless two clusters and a repository have been created.
8) Only one section of the accordion should be able to be opened at one time.
10) When a cluster/repository/migration plan is created listview should open to that section when the modal/wizard closes. **needs fixing for cluster/storage**
12) Wizard step 2 - Validate that a cluster and at least one project have been selected before enabling Next button.
13) Wizard steps 2, 3, 4 - Left align table column headers.

Meeting to discuss these today.

eriknelson:
Added individual issues for the still outstanding issues @vconzola

PR #146, opened 2019-04-29
Collapse other sections when one opened

andyvesel:
Fixes 8 of https://github.com/fusor/mig-ui/issues/97


ibolton336:
We will need to introduce the pf4 accordion component when it is ready for this. Doesnt make sense to modify the data-list to add this. Also, UX said that we are looking at a SPA approach with multiple pages and nav going forward

PR #177, opened 2019-05-09
add initial storage object work

ibolton336:


eriknelson:
🤘

src/app/storage/duck/operations.ts
```diff
@@ -67,12 +110,49 @@ const fetchStorage = () => {
...
+      CoreNamespacedResourceKind.Secret,
+      secretRef.namespace,
+    );
+    refs.push(client.get(secretResource, secretRef.name));
+  });
+
+  return refs;
+}
+
+function groupStorages(migStorages: any[], refs: any[]): any[] {
```
eriknelson: We shoud chat about how to best represent these objects in the front end. I'm curious your opinions on that but you nailed what I did for clusters.

The general opinion I took was, our front end consumes a "cluster". In reality, thats 3 objects. So in the redux tree,  you can get a `cluster.clusterList.{Cluster,MigCluster,Secret}` to retrieve the individual data.

src/client/resources/conversions.ts
```diff
@@ -67,3 +67,63 @@ export function createMigCluster(
     },
   };
 }
+
+export function createMigStorage(
```
eriknelson: +1, I think it's important to have functions that accept the params and spit back the actual kube object.

src/client/resources/mig.ts
```diff
@@ -21,7 +21,7 @@ export class MigResource extends NamespacedResource {
 
 export enum MigResourceKind {
   MigPlan = 'migplans',
-  MigStorage = 'migstorage',
+  MigStorage = 'migstorages',
```
eriknelson: Nice, let's chat about this part of my abstraction, maybe it's a bit weird and could be improved :+1:

PR #182, opened 2019-05-16
Feature/mig api integration

ibolton336:


eriknelson:
Talked with @ibolton336 about this, we need to make sure that we can still run this in local mode with `yarn start` without changing any of the operation code. I'll file an issue for this so we can make sure that gets closed before this gets merged.

eriknelson:
Another blocker is we should get travis up-to-date if there's any work there in master, and then retest this to make sure its up to our standards and is passing CI.

PR #191, opened 2019-05-20
Add tests

andyvesel:
Added basic test & dependencies for UI


andyvesel:
@eriknelson do we need to concern the red travis at this point?

eriknelson:
@andyvesel I'm not sure if we left master in a red state for that last demo. We should start to get more strict about passing travis. I think before we merge this, we should merge 182, and then follow that with a specific PR that fixes the tests to get master back into a green state, and then we can merge this once it also is green. @ibolton336 what do you think about that?

ibolton336:
> @andyvesel I'm not sure if we left master in a red state for that last demo. We should start to get more strict about passing travis. I think before we merge this, we should merge 182, and then follow that with a specific PR that fixes the tests to get master back into a green state, and then we can merge this once it also is green. @ibolton336 what do you think about that?

This sounds good - just realized 182 was still broken. I fixed the tests by just leaving the mock data commented out for now inside kube_store.ts. I also am hardcoding the cluster  namespace list to an empty object for the time being.

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
ibolton336: remove fixtures directory if not in use

cypress/integration/examples/actions.spec.js
```diff
@@ -0,0 +1,272 @@
+/// <reference types="Cypress" />
```
ibolton336: remove examples directory if not in use

cypress/integration/cluster_spec.js
```diff
@@ -0,0 +1,11 @@
+describe("Cluster functionality test", function () {
+  it("Add cluster button opens modal", function () {
+    cy.visit("localhost:9000");
+    cy.get("#add-cluster-btn").then(() => {
+      cy.get("#add-cluster-btn")
+        .first()
+        .click();
+    });
+    cy.contains('Add Cluster')
```
ibolton336: Can you build this out further to run through a full cluster add?

cypress/fixtures/example.json
andyvesel: Removed

cypress/integration/examples/actions.spec.js
andyvesel: Removed

cypress/integration/cluster_spec.js
andyvesel: Added tests for cluster adding with valid data

andyvesel:
@ibolton336 @eriknelson could you review?

cypress/plugins/index.js
```diff
@@ -0,0 +1,17 @@
+// ***********************************************************
```
eriknelson: Is this file required? It's not doing anything functional, we should remove it.

cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
eriknelson: This file is not useful, we should remove it. If it's documentation, it belongs in a README

cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
eriknelson: Also not useful.

src/app/cluster/components/AddClusterModal/AddClusterForm.tsx
```diff
@@ -96,7 +96,7 @@ class WrappedAddClusterForm extends React.Component<any, any> {
             name="url"
             type="text"
             // isValid={!errors.url && touched.url}
-            id="url"
+            id="url-input"
```
eriknelson: +1, we should be better about our IDs and their specificity.

cypress/integration/cluster_spec.js
```diff
@@ -0,0 +1,23 @@
...
+            .type('example.com')
+          cy.get('#token-input')
+            .type('123456')
+          cy.get('#check-connection-btn')
+            .click()
+          cy.get('#submit-cluster-btn')
+            .click()
+        })
+    });
+    cy.contains('Add cluster')
```
ibolton336: need some sort of check to see that the cluster has actually been added. This line was a placeholder until that logic was sorted out.

package.json
```diff
@@ -17,7 +17,8 @@
     "prep-remote": "./scripts/prep-remote.sh",
     "type-check": "tsc --noEmit",
     "type-check:watch": "npm run type-check -- --watch",
-    "test": "jest --rootDir=. --config=config/jest.config.js"
+    "test": "jest --rootDir=. --config=config/jest.config.js",
+    "cypress:open": "cypress open"
```
ibolton336: Like @eriknelson mentioned -> we would benefit from this script running headless instead.

cypress/integration/cluster_spec.js
andyvesel: Added check cluster by name

package.json
andyvesel: Updated scripts, so we have 2 commands to run unit or integration tests and one for running them all

andyvesel:
@eriknelson fyi Cypress's behaviour is to scaffold its default directories (like fixtures, plugins, etc) with every `cypress run`. Its been reported here https://github.com/cypress-io/cypress/issues/501 or here https://github.com/cypress-io/cypress/issues/619
I've disabled these files generating in cypress.json, but when it runs in headed mode on your machine, they will appear again

scripts/test-integration.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+cp ./config/config.dev.json.example ./config/config.dev.json
```
eriknelson: Couple comments on this script:

**Relative vs. Absolute paths**
You're using relative paths, it's safer to use an absolute path that doesn't rely on the current working directly. As an example, the `remote-dev.sh` script [get's the directory that the script is in and sets `$_dir`](https://github.com/fusor/mig-ui/blob/master/scripts/remote-dev.sh#L2). Then you can anchor off of this location to refer to other files by their absolute locations:

```
#!/bin/bash
_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
projectRoot="$_dir/.." # Refers to the root of the git project
cp "$projectRoot/config/config.dev.json.example" "$projectRoot/config/config.dev.json"
# etc..
```
**Overwriting developer config**
` cp ./config/config.dev.json.example ./config/config.dev.json`

This is dangerous because you're going to overwrite any changes a developer has made to their config file if they choose to run this test on their local machine.

You should check to make sure that the file is not present first before copying, something like:
```
if [[ ! -f "$FILE" ]]; then
    echo "$FILE does not exist"
fi
```

So the overall I'd recommend this `test-integration.sh` to address the two issues: https://github.com/eriknelson/mig-ui/blob/andyvesel-add-tests/scripts/test-integration.sh

andyvesel: Thanks for thorough reference, updated.

scripts/test-integration.sh
```diff
@@ -0,0 +1,11 @@
...
+_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+projectRoot="$_dir/.."
+exDevConfigFile="$projectRoot/config/config.dev.json.example"
+devConfigFile="$projectRoot/config/config.dev.json"
+
+if [[ ! -f "$devConfigFile" ]]; then
+  cp "$exDevConfigFile" "$devConfigFile"
+fi
+# TODO: This server dest+port should be parameterized somehow
+./node_modules/.bin/start-server-and-test start http://localhost:9000 './node_modules/.bin/cypress run'
```
eriknelson: You still need to use the `$projectRoot` here: https://github.com/eriknelson/mig-ui/blob/andyvesel-add-tests/scripts/test-integration.sh#L12

cypress/integration/cluster_spec.js
```diff
@@ -0,0 +1,24 @@
...
+    cy.contains('test cluster')
```
ibolton336: Don't need this line anymore.

cypress.json
```diff
@@ -0,0 +1,6 @@
+{
+  "baseUrl": "http://localhost:9000",
+  "video": false,
```
ibolton336: Did we ever figure out if we were storing screenshots/screencaptures of the tests run? I think this option turns off the auto capture of the cypress tests

andyvesel: Yes, I swithed off the video capture. We decided only to take a screenshot after every test passed (I'm gonna add this in future prs). Of course it's not problem to switch video back

Issue #123, opened 2019-04-24
UX issues from 4/25 review


vconzola:
**_This_** means it should be fixed for Summit demo , if possible.

**_1. Masthead background color should be 151515._**
**_2. Status cards at the top should be re-styled per visual design request. See below._**
3. Repositories and plans cards are not showing status.
**_4. Link to add a repository should say "Add repository", not "Add storage"._**
**_5. Empty state for plans should say "Add plan", not "Start plan"._**
6. In the Add cluster and Add repository modals, the "Check connection" button should be disabled until the field content is validated. Otherwise you get a weird interaction where the connection is validated and then you get an error when you click next if a field is empty.
7. Table styling in wizard needs work - column widths, pagination.
8. Wizard step 5 - button label should be Close, not Save and the Back and Cancel buttons should be hidden once the validation is successful.

Status card styling:

![Screen Shot 2019-04-24 at 5 17 02 PM](https://user-images.githubusercontent.com/22823035/56694529-e72ad080-66b4-11e9-8fd0-69aaa43a7ff5.png)

eriknelson:
6. In the Add cluster and Add repository modals, the "Check connection" button should be disabled until the field content is validated. Otherwise you get a weird interaction where the connection is validated and then you get an error when you click next if a field is empty.
8. Wizard step 5 - button label should be Close, not Save and the Back and Cancel buttons should be hidden once the validation is successful.

Think these should be the remaining items from this issue.

PR #195, opened 2019-05-23
Update to use correct repository language

eriknelson:


eriknelson:
Addresses an item from #123

PR #198, opened 2019-05-23
Apply a filter to the source namespaces

eriknelson:


src/app/plan/components/NameSpaceTable.tsx
```diff
@@ -168,8 +168,22 @@ class NamespaceTable extends React.Component<IProps, IState> {
 }
 
 function mapStateToProps(state) {
+  const allSourceClusterNamespaces = state.plan.sourceClusterNamespaces;
+  const filteredSourceClusterNamespaces = allSourceClusterNamespaces.filter(ns => {
```
ibolton336: could move this to a selector. Should open an issue to move all of the derived data to memoized selectors at some point.

Issue #196, opened 2019-05-23
Should checkin the new Logo and remove the yalc instructions


eriknelson:
Logo is public now so we are cleared to check our stuff into the repo and remove the yalc requirements and README instructions.

PR #199, opened 2019-05-23
Add checkconnection component & add logo

ibolton336:


src/app/common/duck/hooks.ts
```diff
@@ -0,0 +1,5 @@
+import { useState } from 'react';
```
eriknelson: Do we want to commit this if it's empty currently?

ibolton336:
closes  #196 #123

src/app/common/components/CheckConnection.tsx
```diff
@@ -0,0 +1,109 @@
...
+          >
+            Add
+          </Button>
+          <Button
+            key="cancel"
+            variant="secondary"
+            onClick={() => this.props.onHandleModalToggle(null)}
+          >
+            Cancel
+          </Button>
```
andyvesel: Does it make sense to put Add and Cancel buttons into the CheckConnection component? They logically more belong to AddClusterForm, rather than to CheckConnection, afaik

PR #218, opened 2019-06-03
Add initial e2e workflow changes

eriknelson:
This PR adds the bulk of the logic to enable real PV discovery, in addition to other fixes in service of driving an initial e2e workflow for a basic stateful move application.

### DIFF
--- a/src/app/plan/duck/operations.ts
+++ b/src/app/plan/duck/operations.ts
@@ -108,14 +108,14 @@ const addPlan = migPlan => {
         migPlan.namespaces,
       );
 
-      let createRes = await client.create(
+      const createRes = await client.create(
         new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
         migPlanObj,
-      )
+      );
 
       dispatch(addPlanSuccess(createRes.data));
 
-      console.debug('Beginning PV polling')
+      console.debug('Beginning PV polling');
 
       const interval = setInterval(async () => {
         const planName = migPlan.planName;
@@ -123,7 +123,7 @@ const addPlan = migPlan => {
         const getRes = await client.get(
           new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
           planName
-        )
+        );
 
         const plan = getRes.data;
         const pvsDiscovered = !!plan.status.conditions.find(c => {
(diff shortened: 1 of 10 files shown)


src/app/plan/duck/operations.ts
```diff
@@ -92,34 +100,74 @@ const addPlan = migPlan => {
...
+
+      dispatch(Creators.updatePlan(latestPlan));
+      const updatedMigPlan = updateMigPlanFromValues(latestPlan, planValues);
+
+      const putRes = await client.put(
+        new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
+        latestPlan.metadata.name,
+        updatedMigPlan,
+      );
+      // TODO: Need some kind of retry logic here in case the resourceVersion
```
eriknelson: This is something we definitely need to add on in a follow up PR, because I've seen it happen if I move through the wizard too fast.

[ISSUE]

eriknelson: This is something we definitely need to add on in a follow up PR, because I've seen it happen if I move through the wizard too fast.

[ISSUE]

src/client/resources/conversions.ts
```diff
@@ -155,13 +154,71 @@ export function createMigPlan(
...
+        name: sourceClusterObj,
         namespace,
       },
+      destMigClusterRef: {
+        name: destinationClusterObj,
+        namespace,
+      },
+      namespaces,
+      //////////////////////////////////////////////////////////////////////////
+      // TODO: Need to rip this out once the controller allows for creation
```
eriknelson: This is required for now, but we'll want to rip this out and test with an updated controller here, so the method continues to make sense.

[ISSUE]

src/client/resources/conversions.ts
```diff
@@ -175,6 +232,11 @@ export function createMigMigration(migID: string, planName: string, namespace: s
         name: planName,
         namespace,
       },
+      // TODO: We need to wire these up to the UI when stage is implemented
+      // and we've more thouroughly tested quiesce/stage combos in conjunction
+      // with the controller and velero.
+      quiescePods: true,
```
eriknelson: [ISSUE]

src/app/plan/duck/operations.ts
eriknelson: https://github.com/fusor/mig-ui/issues/222

src/client/resources/conversions.ts
eriknelson: https://github.com/fusor/mig-ui/issues/227

src/client/resources/conversions.ts
eriknelson: #224, #225

PR #219, opened 2019-06-04
Fix wizard persist state bug

### LINKED ISSUE #122: Wizard opening on Step 5
After creating a plan, if user selects Add plan button the wizard opens at step 5 with validation running and then shows validation success with no plan name.
Screen Shot 2019-04-24 at 3 49 04 PM

### LINKED ISSUE #239: Address next button logic bugs
Currently, the wizard is behaving not as expected when navigating backwards . 

Also, there needs to be a few different validation paths for stateless app migrations.

### LINKED ISSUE #240: Move wizard modal state to redux
Several bugs are popping up related to how we are using state across multiple components within the wizard modal. Moving the isWizardOpen state, enableNextState, & any other logic that is being changed over multiple page components should help resolve these bugs.

### LINKED ISSUE #246: Dropdown values resetting when navigating away from form inside wizard
Dropdown should display the current value when revisiting the form.

### LINKED ISSUE #247: Duplicate plan creation on navigation
Need to sort out the add plan on navigate functionality. If a user navigates backwards & then moves forward again, this causes duplicate plans to be created. One suggestion would be to disable the back button after the first step & move the plan creation to the first step.

ibolton336:
Closes #122 
Closes #246 
Closes #247 
Closes #239 
Closes #240


src/app/plan/components/Wizard/WizardComponent.tsx
```diff
@@ -0,0 +1,149 @@
...
+    }
+  };
+  const handleClose = () => {
+    props.onHandleClose();
+    props.resetForm();
+  };
+
+  return (
+    <React.Fragment>
+      {props.isOpen && (
```
ibolton336: @eriknelson the fix ^^^^^^^

src/app/home/components/DataList/Clusters/ClusterContent.tsx
```diff
@@ -33,7 +36,14 @@ const ClusterContent = ({
       ) : (
         <Flex alignItems="center" justifyContent="center">
           <Box>
-            <DataListEmptyState type="cluster" />
```
ibolton336: did away with the DataListEmptyState component b/c it wasnt really reducing any boilerplate

src/app/plan/components/Wizard/VolumesForm.tsx
```diff
@@ -26,29 +27,33 @@ class VolumesForm extends React.Component<any> {
...
-      return p.metadata.name === values.planName;
-    });
+      const currentPlan = plans.find(p => {
+        return p.metadata.name === values.planName;
+      });
 
-    if(!currentPlan.status) {
-      return;
-    }
+      if (currentPlan) {
```
ibolton336: Fixed a bug causing the app to crash when state was between renders & currentPlan was not defined yet

src/app/plan/components/Wizard/VolumesForm.tsx
```diff
@@ -62,7 +67,22 @@ class VolumesForm extends React.Component<any> {
     const currentPlan = plans.find(p => {
       return p.metadata.name === values.planName;
     });
-
+    if (this.props.isPVError) {
```
ibolton336: added this component level error state to handle the case of no PVs showing up after 3 checks of the interval. When the isPVError flag is checked, the volumes form will show this error. Need to add redux actions & state to encapsulate validation so that this error state enables the user to go to next page.

src/app/plan/duck/operations.ts
```diff
@@ -127,8 +128,15 @@ const addPlan = migPlan => {
...
       console.debug('Beginning PV polling');
-
+      let timesRun = 0;
       const interval = setInterval(async () => {
+        timesRun += 1;
+        if (timesRun === 3) {
+          clearInterval(interval);
+          dispatch(commonOperations.alertErrorTimeout('No PVs found'));
+          dispatch(pvFetchFailure());
+        }
```
ibolton336: this is where we would trigger a validation check for this wizard step & update the wizard next enable value.

src/app/plan/duck/operations.ts
```diff
@@ -127,8 +128,15 @@ const addPlan = migPlan => {
       dispatch(addPlanSuccess(createRes.data));
 
       console.debug('Beginning PV polling');
-
+      let timesRun = 0;
       const interval = setInterval(async () => {
+        timesRun += 1;
+        if (timesRun === 3) {
```
ibolton336: added a variable to keep track of the interval.

src/app/plan/duck/operations.ts
```diff
@@ -137,14 +145,17 @@ const addPlan = migPlan => {
...
 
         const plan = getRes.data;
-        const pvsDiscovered = !!plan.status.conditions.find(c => {
-          return c.type === PvsDiscoveredType;
-        });
-
-        if (pvsDiscovered) {
-          console.debug('Discovered PVs, clearing interaval.');
-          clearInterval(interval);
+        if (plan.status) {
```
ibolton336: wrapped this in a conditional to allow this to run in local dev mode also.

src/app/plan/components/Wizard/VolumesTable.tsx
```diff
@@ -49,46 +57,124 @@ class VolumesTable extends React.Component<any, any> {
...
-    return discoveredPersistentVolumes.map(planVolume => {
-      let pvAction = 'copy'; // Default to copy
-      if(this.props.values.persistentVolumes.length !== 0) {
-        const rowVal = this.props.values.persistentVolumes.find(v => v.name === planVolume.name);
-        pvAction = rowVal.type;
+      // No PVs discovered to be in use. This is normal for stateless cloud apps.
+      if (!discoveredPersistentVolumes) {
+        return [];
       }
+      if (this.props.values.prersistentVolumes) {
```
eriknelson: Some logic is broken here unless there's a consistent typo: `prersistentVolumes`

src/app/plan/duck/operations.ts
```diff
@@ -126,9 +129,19 @@ const addPlan = migPlan => {
...
       dispatch(addPlanSuccess(createRes.data));
 
-      console.debug('Beginning PV polling');
-
+      let timesRun = 0;
       const interval = setInterval(async () => {
+        dispatch(pvFetchRequest());
+        timesRun += 1;
+        // TODO: replace timesRun with poller class
+        if (timesRun === 3) {
```
eriknelson: I don't remember what the time interval is, but we should bump this up so that it's a total of something like 30 seconds. I've seen it take on the order of 10. Also, we should put the times run condition in a const var that's defined next to the polling interval var, so that is easy to tweak:

```
const watchInterval = 5000; // 5 seconds
const watchTicks = 12; // Check 12 times, so timeout is 60 secs

//...
         if (timesRun === watchTicks) {
```

src/app/plan/components/Wizard/VolumesTable.tsx
ibolton336: updated this to work now. Nice catch

src/app/plan/duck/operations.ts
ibolton336: Updating to use timesRun === 12. The times run is set outside the interval function using let so that it can be updated on every tick of the interval. Definitely think this should be scrapped asap for the poller class though.

Issue #62, opened 2019-04-05
Re-evaluate using reduxsauce due to type concerns


eriknelson:
Currently fixing a bug (#61) that's a direct result of us opting out of typescript type checking. Basically all the error dispatches were passing in an Error type, when the reducer was expecting a string. Is it possible to continue using `reduxsauce` to generate our action creators and make them strongly typed? If not, I think we should re-evaluate how we want to write our action creators, and possibly drop reduxsauce.

ibolton336:
I'm ok with dropping redux sauce

eriknelson:
Honestly I like the convenience it offers a lot, it'd be nice if there was a typed approach available.

PR #221, opened 2019-06-04
refactor alerts to not use redux sauce & add pf4 auto hiding alert

### LINKED ISSUE #214: Success and failure toast notifications
According to https://github.com/fusor/mig-ui/pull/211#discussion_r289669972 we need to add success and failure [notifications](https://www.patternfly.org/pattern-library/communication/toast-notifications/) for such events as add/remove clusters, storage, plans, etc

ibolton336:
Closes #214 #62


andyvesel:
@ibolton336 I think we also need to add success notifications here for remove cluster or storage, b/c for now, when delete a cluster/storage, nothing happens on the screen and it needs to manually refresh the page.

ibolton336:
> @ibolton336 I think we also need to add success notifications here for remove cluster or storage, b/c for now, when delete a cluster/storage, nothing happens on the screen and it needs to manually refresh the page.

@andyvesel This PR will need to go in so that we can use the updated alert notification code on johns PR.

src/app/common/duck/actions.ts
```diff
@@ -1,9 +1,13 @@
...
+export const ALERT_CLEAR = 'ALERT_CLEAR';
 
-const { Creators, Types } = createActions({
-  alertSuccess: ['success'],
-  alertError: ['alertMessage'],
-  alertClear: [],
-});
-
-export { Creators, Types };
+export function alertSuccess(text) {
```
eriknelson: Can we enforce with TS that `text` here can only be a string? We've had multiple issues with `errors` getting shoved into these methods and it blowing up at runtime. It'd be great if we could lean on TS more to check this stuff when possible at compile time.

src/app/common/duck/operations.ts
```diff
@@ -1,9 +1,31 @@
-import { Creators } from './actions';
-const alertSuccess = Creators.alertSuccess;
-const alertError = Creators.alertError;
-const alertClear = Creators.alertClear;
+import { alertSuccess, alertError, alertClear } from './actions';
+const alertErrorTimeout = message => {
```
eriknelson: The arguments to these operations should also only accept TS strings, i.e. `message: String`.

PR #250, opened 2019-06-11
Add some copy changes

### LINKED ISSUE #210: Remove text of 'AWS S3' and replace with a more generic 'S3' in add Replication Repository form
See below 'AWS S3 Bucket Name' and 'AWS S3 Bucket Region'

Screen Shot 2019-05-31 at 8 18 17 PM

eriknelson:
Closes #210

* Keep add cluster modal title consistent with add storage modal. Both
should be "Add X".
* Remove "AWS" from the labels on the add storage form, since they will
not always actually be AWS

### DIFF
--- a/src/app/cluster/components/AddClusterModal/index.tsx
+++ b/src/app/cluster/components/AddClusterModal/index.tsx
@@ -27,7 +27,7 @@ class AddClusterModal extends React.Component<any, any> {
     const { checkConnection, connectionState, name, url, token, mode } = this.props;
 
     return (
-      <Modal isSmall isOpen={this.props.isOpen} onClose={this.handleClose} title="Add Cluster">
+      <Modal isSmall isOpen={this.props.isOpen} onClose={this.handleClose} title="Cluster">
         <AddClusterForm
           connectionState={connectionState}
           onHandleModalToggle={this.handleClose}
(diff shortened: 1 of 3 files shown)


src/app/cluster/components/AddClusterModal/index.tsx
```diff
@@ -27,7 +27,7 @@ class AddClusterModal extends React.Component<any, any> {
     const { checkConnection, connectionState, name, url, token, mode } = this.props;
 
     return (
-      <Modal isSmall isOpen={this.props.isOpen} onClose={this.handleClose} title="Cluster">
+      <Modal isSmall isOpen={this.props.isOpen} onClose={this.handleClose} title="Add Cluster">
```
ibolton336: I think John removes the add because the same modal is used for editing @jwmatthews ?

eriknelson: Ah that makes sense. Do we need to remove "Add" from "Add Storage" then? I thought it was odd that they were inconsistent.

ibolton336:
Yeah that’s a good call. We should update both. Is it worth having dynamic text based on the action selected?

eriknelson:
I don't think it's worth dynamic tech. The components are specific to their type anyway.

eriknelson:
The decision was to make these agnostic until we can add in dynamic titles that reflect the action of the modal. Filed an issue to cover the dynamic titles: #257

PR #265, opened 2019-06-13
Add Mig UI production deployment

### LINKED ISSUE #258: UI should be deployable and run independently from a cluster
A user should be able to deploy the UI easily into a cluster using some kind of installation script, whether that's ansible or bash. We need a way to provide configuration data to the front end via `migMeta` (things like the clusterAPI address, where to create secrets, what namespace the API objects should be created in, etc.). Webpack delivers this data to the front-end by reading a local `config.dev.json` file currently, but inside of a cluster, this is going to require a lightweight server that has a few responsibilities:

* Serve static, like the production js bundle, fonts, images, etc.
* When the UI is requested, dynamically read configuration data from something like a `ConfigMap`, encode that data and anything else that the UI would need to know into a base64 encoded `migMeta` object, and then serve the apps html template with that encoded blob + the script tag to load the app.

Overall, this issue encompasses:
* Writing the mig-ui-server
* Ensuring the production build is working (Closes #154)
* Building a mig-ui image that packages everything we need and runs the mig-ui-server
* Working through any integration issues that crop up

eriknelson:
Closes #258

Note, this requires folks to copy over the latest `config.dev.json` from from the `config.dev.json.example` file to pickup the configurable `oauthConfigSecret`.


eriknelson:
Deployment README found at `deploy/README.md`

eriknelson:
@jwmatthews Please review the `deploy/README.md` in particular.

src/app/auth/LoginComponent.tsx
```diff
@@ -29,21 +29,9 @@ class LoginComponent extends React.Component<IProps> {
...
-      ////////////////////////////////////////////////////////////
-      // TODO: Currently using an empty secret value here since we
-      // are strictly a public client
-      // The correct thing to do here is to implement PKCE, which is
-      // what oc uses and what is specfically implemented for public
-      // clients: https://tools.ietf.org/html/rfc7636
-      // For now, using a temporarily hardcoded secret until PKCE can be
-      // implemented. (migrations.openshift.io | base64)
-      //
-      const loginSecret = 'bWlncmF0aW9ucy5vcGVuc2hpZnQuaW8K';
```
eriknelson: Important that this is dynamically generated.

src/index.tsx
```diff
@@ -18,7 +18,7 @@ const onBeforeLift = () => {
 };
 // Some amount of meta data is delivered to the app by the server
 /* tslint:disable:no-string-literal */
-const migMeta = window['_mig_meta'];
+const migMeta = JSON.parse(atob(window['_mig_meta']));
```
eriknelson: This probably should have been a b64 value from the beginning. Solves a lot of problems.

public/index.ejs
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en-US">
+
+<head>
+  <meta charset="utf-8" />
+  <title>Cluster Application Migration Tool</title>
+  <meta id="appName" name="application-name" content="Mig UI" />
+  <!-- <link rel="icon" type="image/png" href="/assets/favicon.ico"> -->
+  <!-- <link rel="stylesheet" href="../node_modules/@patternfly/react-core/dist/styles/base.css" /> -->
```
eriknelson: @ibolton336 I think these are what I need help with. I think I need to copy this file out of the node_modules and expose it from the file server so we load it prior to the app bundle. I'm not sure how to do that in the webpack production build.

eriknelson:
This PR is largely complete; the issue I'm still seeing is some bad styles in a production build. I suspect it's due to the linked comment: https://github.com/fusor/mig-ui/pull/265#discussion_r295575348

I think I need a webpack plugin for the production build that picks up the base patternfly styles and drops them into `dist`, so they end up in the staticroot of the image and get loaded prior to the app bundle.

PR #267, opened 2019-06-14
First cut at automating fetch of mocked data from a server

jwmatthews:
@andyvesel this is what I'm thinking for how we gather and structure the data.

Big pieces are instead of manually fetching each json snippet, we should automate all of the fetching.
The data is likely to change often, best to have a script fetch good data from known clusters and store it.

This way we can treat all of the generated data under './mocked_data/data' as replaceable.

Intent is that we will add functionality into kube_store.ts that will read all data under './mocked_data/data' and use the data from disk to serve requests.   

I added a 'TIMESTAMP' file which tracks last time the data was gathered as well, so we can use that to compare against what is in localStorage, if the data from our repo is new then thinking we'd wipe data form localStorage and load from disk.


jwmatthews:
Note, I am intentionally committing data to 'mocked_data/data'.
The data committed is intended to be wiped out and re-fetched frequently, hence a TIMESTAMP is also created to track last time data was gathered.

Thought is that no one will make manual changes to anything under 'mocked_data/data'.

The expected usage is we will stand up a working environment with real data.
Then we will run 'gather_sample_data.sh', it will talk to the configured HOST and pull down all of the sample json data to use for mocks.

I am also intending to ensure we remove any sensitive aws credentials, so added a section to remove aws credentials and replace with dummy data.

PR #268, opened 2019-06-15
Mocking data for local usage

jwmatthews:
This PR changes 'KubeStore' to use localStorage for mocked data.

Basic concept is that KubeStore will load mocked data from a local json file into localStorage and will then serve requests from the data in localStorage.

I took a stab at manually creating the format for the mocked data, have it working with entities for MigCluster only.  This approach is intended to support data from multiple hosts, will need to flesh out more data to confirm implementation supports.

Next steps:
- [x] Need to rework the fetching of data from a host to automatically populate the mocked data.
- [ ] Add support for 'put' in client.mock. 
- [ ] Add support for 'put' in KubeStore, it should be able to modify data in localStorage,  I think this will need work per Resource type
- [ ] Add support for 'patch' in KubeStore, it should be able to modify data in localStorage, also work per resource type
- [ ] Test with multiple hosts
- [ ] Need to set up a good 2 cluster env with successful migrations run to pull data and update mock data
- [ ] Clean up code and remove unused older mocked_data and KubeStore db references.


jwmatthews:
@andyvesel here's an approach for modifying KubeStore to read in the mocked data and populate localStorage.   

Please take a look at this approach and we can work together on finishing the pieces during the week.  
To support the upcoming webinar need to get the mocked data complete and updated before end of the week.

jwmatthews:
@andyvesel updated PR so at this point we can run:
'yarn mocks' it will talk to the k8s cluster referenced in the kubeconfig and will pull down the needed data.

I've updated KubeStore to read the mocked data.
Get and List are working.

Thinking might be good to merge this PR after it's been reviewed and follow up with new PRs that add in support for Put/Patch

andyvesel:
@jwmatthews ack, thanks for good example of this.

scripts/gather-mocked-data.js
```diff
@@ -0,0 +1,186 @@
...
+      });
+  });
+
+  return all.reduce((promiseChain, currentTask) => {
+    return promiseChain.then(chainResults => {
+      return currentTask.then(currentResult => {
+        if (!chainResults[currentResult.key]) {
+          chainResults[currentResult.key] = {};
+        }
+        chainResults[currentResult.key][currentResult.name] = currentResult.data;
```
jwmatthews: This data needs to be sanitized for any aws access keys to ensure we don't accidentally commit real data.

jwmatthews:
Please hold off on merging until we are sure that the aws access credentials are sanitized and never committed.  At moment this is all test/dummy data so nothing has leaked, but this needs to be handled for future before merging.

scripts/gather-mocked-data.js
jwmatthews: Resolved

jwmatthews:
Good to merge, the AWS key issue has been addressed

Support for get/list is present.

Next up is put/patch we will add that in separate PRs

ibolton336:
image

This is looking good - I am wondering if there are any plans to mock out the status on the plan? even if we could just get a completed status immediately on plan creation, that would help when working on UI specific css/features locally.

jwmatthews:
> Looking good - Can we get a set of initial data that would allow a dev to jump into ui development immediately? I am missing namespaces on the mocked out target cluster. Also, some status mocking for plan creation would be great.

Plan is to add more data, this PR is just an initial step.

Only has data for seeing clusters/repositories.

PR #282, opened 2019-06-26
Fix deployed base styles

### LINKED ISSUE #275: Production UI not using pf4-react-core base.css files correctly

eriknelson:
Closes #275

This PR adds some build improvements that breaks out the css from the JS payload into its own file for production builds.

WIP since there are still broken PF styles in a number of places that we're trying to work through.


eriknelson:
This PR separates out the css styles into their own file. For some reason, when doing this in an in-cluster production setting, the **JS bundle** takes almost 30 seconds to load compared to when we leave the CSS as part of the JS bundle like we do today, which loads in 5-10s. Seems totally counter-intuitive, but it's demonstrable.

Overall we have a bunch of optimizations to do in this area, but I'm going to close this in favor of the faster option which is what we have today. We can reopen this when we optimize if we want to do parallel css loads.

PR #283, opened 2019-06-26
New switch for namespace removal (namespace creation switch removed)

pranavgaikwad:
This enables the user to select when they want to delete the namespace. The namespace is always created when it does not exist.

Signed-off-by: Pranav Gaikwad <pgaikwad@redhat.com>


deploy/destroy.sh
```diff
@@ -6,6 +6,10 @@ if [[ "$HOSTAPI" == "" ]]; then
   exit 1
 fi
 
+if [[ "$CLEANUP" == "" ]]; then
+  CLEANUP="false"
+fi
+
 pushd $_dir/deploy
-ansible-playbook -e mig_ui_cluster_api_endpoint="$HOSTAPI" -e state=absent playbook.yaml
+ansible-playbook -e mig_ui_cluster_api_endpoint="$HOSTAPI" -e state=absent playbook.yaml --extra-vars='{"mig_ui_remove_namespace": '$CLEANUP'}'
```
eriknelson: You should use the same pattern as the other `-e` vars here.

deploy/roles/mig-ui/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
   k8s:
     state: '{{ state }}'
     definition: "{{ lookup('template', 'mig-ui.namespace.yaml') | from_yaml }}"
-  when: mig_ui_create_namespace
+  when: mig_ui_remove_namespace or state == 'present'
```
eriknelson: I don't think this logic is semantically correct. It probably works as intended, but it's going to run this task **always** when `mig_ui_remove_namespace` is true, which isn't really what we want.

We want to "delete this namespace when state absent and the user has specifically indicated they want to remove the namespace, OR when the state is present to ensure the namespace exists. I would change this to:

`(mig_ui_remove_namespace and state == 'absent') or state == 'present'`

deploy/destroy.sh
pranavgaikwad: the problem with using -e is that all variables passed with -e are parsed as strings. while the CLEANUP is a boolean. perhaps, you could suggest a better way to pass boolean value

eriknelson:
CC: @jmontleon

Closing this since the current plan is to move this role into a centralized deployment repo hosing the entire Mig project's deployment and operator scripts. Since the namespace management is really the domain of that project and may change depending on needs there, I'm going to close this since it's made obsolete by that change.

PR #289, opened 2019-06-28
Implement stage & add status polling for loading indicator

### LINKED ISSUE #225: Need to implement Stage
Stage should create a migration almost exactly the same way as Migrate does, but set the `stage` field on the posted `MigMigration` to true. I would think these two should share almost identical code, so this could almost just be a flag.

https://github.com/fusor/mig-ui/pull/218#discussion_r290807746

### LINKED ISSUE #238: Util for handling controller conditions
We need to have a utility function that translates the controller conditions into something consumable by the ui.

### LINKED ISSUE #279: Provide feedback acknowledging the migration was accepted
With 277, the UI doesn't provide any feedback to the user to ack that the Migration has been accepted until the controller has actually started to run it. We should do something to indicate this right after the user has started it.

### LINKED ISSUE #288: Console is filled with sourcemap load failures
One of the recently merged PRs broke sourcemap loads, so the console is completely filled with load errors. We should fix this for both dev and prod modes.

### LINKED ISSUE #290: Update plan status counts with real data
Pull status from plan conditions & reflect this in the status card

ibolton336:
Closes #288 
Closes #238 
Closes #225 
Closes #290 
Closes #279


src/app/home/components/DataList/Plans/PlanActions.tsx
```diff
@@ -1,79 +1,79 @@
-import React, { useState } from 'react';
+/** @jsx jsx */
```
eriknelson: What's this for?

src/app/home/components/DataList/Plans/PlanStatus.tsx
```diff
@@ -1,59 +1,47 @@
...
-      }
-      if (hasSucceededMigration) {
-        status.text = 'Succeeded';
-      }
-      if (hasClosedCondition) {
-        status.text = 'Closed';
-      }
+  const getStatus = () => {
+    const status = { text: 'Waiting for status...' };
+    if (hasReadyCondition || !hasPrevMigrations) {
```
eriknelson: The way this ladder of if statements is structured, it's going to set the text in order of precedence from top to bottom. so if more than one applies, the lowest one is going to be the one that's set. I don't know if that's the intention or not, but just wanted to call that out. If we want to short circuit the first match, a switch with breaks might be a better choice.

src/app/plan/duck/operations.ts
```diff
@@ -30,82 +30,170 @@ const pvFetchRequest = Creators.pvFetchRequest;
...
+        true
+      );
+      const migMigrationResource = new MigResource(MigResourceKind.MigMigration, migMeta.namespace);
+
+      //created migration response object
+      const createMigRes = await client.create(migMigrationResource, migMigrationObj);
+      const migrationListResponse = await client.list(migMigrationResource);
+      const groupedPlan = groupPlan(plan, migrationListResponse);
+
+      const callbackFunction = response => {
```
eriknelson: I might give this a more descriptive name to indicate what its logic is actually doing.

src/app/plan/duck/operations.ts
```diff
@@ -30,82 +30,170 @@ const pvFetchRequest = Creators.pvFetchRequest;
...
-            i => i.kind === 'MigMigration' && i.spec.migPlanRef.name === plan.MigPlan.metadata.name
-          );
-
-          fullPlan['Migrations'] = sortMigrations(matchingMigrations);
-        } else {
-          fullPlan['Migrations'] = [];
+      const migrationListResponse = await client.list(migMigrationResource);
+      const groupedPlan = groupPlan(plan, migrationListResponse);
+
+      const callbackFunction = response => {
```
eriknelson: Same comment here as above. I wonder if the logic in this is shared between the two callback functions at all and could be pulled into another fn?

src/app/common/duck/sagas.ts
```diff
@@ -19,7 +19,7 @@ function* poll(action) {
       // Make the API call
       stats.fetching = true;
       params.onStatsChange(stats);
-      const response = yield call(params.asyncFetch);
+      const response = yield call(params.asyncFetch, null);
```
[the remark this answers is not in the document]
ibolton336: This can actually be deleted. I was debugging an issue and i passed this in to try to fix it. Will remove.

ibolton336:
> Seems like a lot of the generic polling names were updated to be specific to `DataList`, I assume because we're polling other things now as well? I think my next PRs will be related to Check Connection and I'd like to dive into the saga stuff for that.

Yeah I was running into some issues having that one polling function be the one size fits all approach. There is still some room for improvement & we could iterate on the sagas further for sure. The primary issue i was running into with reusing that generator I think was my specific combination of thunks/sagas/redux sauce. Right now, I am using the sagas in "common" as the main background task for datalist polling. I added another saga to the plans directory that is shared between runMigration/runStage for polling the status of a specific migration/stage. When an error/success condition occurs within the callback function, the generator is paused/stopped & the redux store is updated with isStaging/isMigrating to false. This was needed as a way to find out when specific migrations/stages were completing/starting so that the ui could be updated precisely when the user clicks stage/migrate rather than just waiting for the status from the background datalist poll. @eriknelson 
> 
> Just a few other comments, otherwise, looks good. It'd be great if we could clean up some of those crazy nested conditionals, I'm not sure how to really do that given all the branching logic that's actually necessary.

src/app/plan/duck/operations.ts
ibolton336: yeah definitely some room for refactoring here. I will see what I can do

src/app/plan/duck/operations.ts
ibolton336: :+1:

src/app/home/components/DataList/Plans/PlanStatus.tsx
ibolton336: Yeah this was something I was meaning to ask you about. The logic gets a bit hairy here so I did order them in a specific way. For example, If there is a ready condition and there are migrations running, i want the running status to be shown rather than ready even though the plan is still ready.

src/app/home/components/DataList/Plans/PlanActions.tsx
ibolton336: This is used in lieu of the emotion babel plugin to get the css prop to work with emotion.

ibolton336: https://emotion.sh/docs/css-prop#jsx-pragma

src/app/home/components/DataList/Plans/PlanStatus.tsx
eriknelson: Cool, just wanted to make sure there wasn't unintended behavior here. At some point it'd be cool to spend some time refactoring this so it's a little more maintainable and cleaner, but I think that's out of scope for this PR. Worth an issue.

ibolton336: #291 filed

src/app/plan/duck/operations.ts
```diff
@@ -30,85 +30,168 @@ const pvFetchRequest = Creators.pvFetchRequest;
...
+
+      //created migration response object
+      const createMigRes = await client.create(migMigrationResource, migMigrationObj);
+      const migrationListResponse = await client.list(migMigrationResource);
+      const groupedPlan = groupPlan(plan, migrationListResponse);
+
+      const statusPollingCallback = response => {
+        if (response && response.isSuccessful === true) {
+          const isStage = true;
+          return checkMigrationStatus(dispatch, response, createMigRes, isStage);
```
ibolton336: @eriknelson extracted the shared logic out to this function

eriknelson:
Large PR so did my best to highlight anything I happened to see. Also pulled and ran a migration with success. Definitely a UX improvement as far as I can tell. Thank you @ibolton336!

PR #294, opened 2019-07-01
Patch request

andyvesel:
Added patch request for mocked data

### DIFF
--- a/src/client/kube_store.test.ts
+++ b/src/client/kube_store.test.ts
@@ -1,6 +1,7 @@
 import _ from 'lodash';
 import KubeStore from './kube_store';
 import { MigResource, MigResourceKind } from './resources';
+import mocked_data from './kube_store/mocked_data/';
 
 const examplePlan = {
   apiVersion: 'migration.openshift.io/v1alpha1',
@@ -36,6 +37,7 @@ const examplePlan = {
 const testNs = 'test-ns';
 const planName = 'my-plan';
 const expectedGvk = 'migration.openshift.io/v1alpha1/migplans';
+const clusterName = 'foo';
 
 test('Test NamespacedResource setResource', () => {
   const migResource = new MigResource(MigResourceKind.MigPlan, testNs);
@@ -53,4 +55,14 @@ test('Test NamespacedResource setResource', () => {
 
   store.setResource(migResource, planName, examplePlan);
   expect(_.isEqual(expected, store.db)).toBe(true);
+});
+
+test('Test NamespacedResource patchResource', () => {
+  const migResource = new MigResource(MigResourceKind.MigCluster, 'mig');
+  const patch = { 'spec': { 'patch': { 'test': 'foo' } } };
+  const store = new KubeStore('_host');
+  const expected = patch['spec']['patch'];
+  const patchedObject = store.patchResource(migResource, clusterName, patch);
+
+  expect(_.isEqual(expected, patchedObject['spec']['patch'])).toBe(true);
 });
--- a/src/client/kube_store/index.ts
+++ b/src/client/kube_store/index.ts
@@ -35,6 +35,40 @@ export default class KubeStore {
     return d['clusters'][this.clusterName];
   }
 
+  private determineKey(resource: KubeResource, name: string): string {
+      // Determine the url path as the 'key' for Resource or List of Resources
+      // 2 Main Types of Resources
+      // - NamespacedResource
+      // - ClusterResource
+      // Next 2 distinctions,  'Core' or Extension.
+      //  Core - are the main types built in k8s
+      //  If it's an instance of CoreNamespacedResource or CoreClusterResource
+      // Extensions are everything else.
+
+    let key: string = '';
+
+    if (resource instanceof NamespacedResource) {
+      const namespacedResource = resource as NamespacedResource;
+      if (resource instanceof CoreNamespacedResource) {
+        // Core Namespaced
+        key = `api/${resource.gvk().version}/namespaces/${namespacedResource.namespace}/${resource.gvk().kindPlural}`;
+      } else {
+        // Extension Namespaced
+        key = `apis/${resource.gvk().group}/${resource.gvk().version}` +
+          `/namespaces/${namespacedResource.namespace}/${resource.gvk().kindPlural}`;
+      }
+    } else {
+      if (resource instanceof CoreClusterResource) {
+        // Core ClusterResource
+        key = `api/${resource.gvk().version}/namespaces/${resource.gvk().kindPlural}`;
+      } else {
+        // Extension ClusterResource
+        key = `apis/${resource.gvk().group}/${resource.gvk().version}/${resource.gvk().kindPlural}`;
+      }
+    }
+    return key;
+  }
+
   private ensureDataLoaded() {
     let localData = JSON.parse(localStorage.getItem(localStorageMockedDataKey));
     if ((!localData) || (!localData.TIME_STAMP) || (localData.TIME_STAMP < mocked_data.TIME_STAMP)) {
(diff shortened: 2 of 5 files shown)


src/client/kube_store/index.ts
```diff
@@ -233,6 +48,47 @@ export default class KubeStore {
     return this.instance || (this.instance = new this(''));
   }
 
+  public patchResource(resource: KubeResource, name: string, patch: object): object {
+    let source: object = {};
+    let key: string = '';
+    const myPatch: object = {'foo':'bar'};
```
eriknelson: Don't think we want dummy data here.

eriknelson:
Needs rebase

src/client/kube_store/index.ts
```diff
@@ -1,9 +1,10 @@
 import _ from 'lodash';
 import { NamespacedResource, ClusterResource, KubeResource,
   CoreNamespacedResource, CoreClusterResource, } from '../resources';
+import { ClusterClient } from '../client';
```
jwmatthews: I suspect this import isn't needed.

src/client/kube_store/index.ts
jwmatthews: +1 remove this and just use 'patch' object that is passed in

src/client/kube_store/index.ts
```diff
@@ -233,6 +48,47 @@ export default class KubeStore {
...
+        }
+      } else {
+        // Extension ClusterResource
+        key = `apis/${resource.gvk().group}/${resource.gvk().version}/${resource.gvk().kindPlural}`;
+        if (this.data()[key]) {
+          source = this.data()[key][name];
+        }
+      }
+    }
+    const patchedObject = JsonMergePatch.apply(source, myPatch);
```
jwmatthews: assume this should be 

const patchedObject = JsonMergePatch.apply(source, patch);

src/client/kube_store/index.ts
```diff
@@ -233,6 +48,47 @@ export default class KubeStore {
     return this.instance || (this.instance = new this(''));
   }
 
+  public patchResource(resource: KubeResource, name: string, patch: object): object {
+    let source: object = {};
+    let key: string = '';
+    const myPatch: object = {'foo':'bar'};
+    if (resource instanceof NamespacedResource) {
```
jwmatthews: As we discussed, later when we know this function is working, we can explore refactoring it so that determining the key is in a different function which can be reused.

src/client/kube_store/index.ts
```diff
@@ -233,6 +47,48 @@ export default class KubeStore {
...
+      } else {
+        // Extension ClusterResource
+        key = `apis/${resource.gvk().group}/${resource.gvk().version}/${resource.gvk().kindPlural}`;
+        if (this.data()[key]) {
+          result = this.data()[key][name];
+        }
+      }
+    }
+
+    const patchedObject = JsonMergePatch.apply(resource, patch);
```
jwmatthews: Patching 'resource' is incorrect.
In this context, 'resource' is an object that just tracks the GVK metadata, it does not have the actual data.

(This may be confusing as the name 'Resource' is used in k8s and typically would refer to object data, in this case we are using the class definition 'Resource' to track metadata associated with the GVK.  'Resource' in this context is essentially a class that helps us to identify the URL of the Resource on k8s API)

What you want to do is to apply the json merge patch to the data.

You want to:
1. Extract the desired data (you use the resource and it's gvk methods to form a key so you can lookup the specific data you want from localStorage).  
2. Apply the merge patch to the actual data which is in the store for this specific object

One issue in the method way it is written is that you are not getting the data.

You have a block above which is doing a lookup on this.data()[key][name], and putting the data in 'result'... yet you are not using 'result'.

I would have expected to see:
const patchedObject = JsonMergePatch.apply(result, patch);

src/client/kube_store/index.ts
jwmatthews: Looks like you are ignoring this entire block and not using the 'result'

The 'result' is the important part of how we do the lookup of mocked data

src/client/kube_store/index.ts
```diff
@@ -233,6 +47,48 @@ export default class KubeStore {
...
+    localStorage[key] = patchedObject;
```
jwmatthews: This is wrong.

As we discussed, we are storing the entire mocked data as a single key in localStorage.
So we have a single key for ALL of our data.

This method as written implies that we are storing individual elements in localStorage with a primary key, we are not doing that.  We are using a series of nested keys.

src/client/kube_store/index.ts
```diff
@@ -233,6 +47,48 @@ export default class KubeStore {
...
+// console.log(patchedObject)
+    return resource;
```
jwmatthews: We wouldn't return 'resource'.

Perhaps we'd return the patchedObject or maybe we just return a success or error condition.

src/client/kube_store.test.ts
```diff
@@ -54,3 +56,24 @@ test('Test NamespacedResource setResource', () => {
...
+  const patch = {'test':'foo'};
+
+  const expected = {
+
+      namespace: testNs,
+        _gvk:
+          { group: 'migration.openshift.io',
+            version: 'v1alpha1',
+            kindPlural: 'migclusters' },
+          test: 'foo',
```
jwmatthews: This doesn't line up to what we'd normally see.

We'd expect the data to be structured under a 'spec': {}
So would expect something more like
{ 'spec': {'test': 'foo'}}

src/client/kube_store.test.ts
```diff
@@ -54,3 +56,24 @@ test('Test NamespacedResource setResource', () => {
...
+    };
+
+  const patchedObject = store.patchResource(migResource, clusterName, patch);
+  localStorage.clear();
```
jwmatthews: I don't follow why we'd clear localStorage

src/client/kube_store.test.ts
```diff
@@ -54,3 +56,24 @@ test('Test NamespacedResource setResource', () => {
...
+  localStorage.setItem('test', JSON.stringify(patchedObject))
+  expect(patchedObject).toEqual(expected);
```
jwmatthews: We want to test how the patch command works.
So the workflow would be something like:

1) Determine some piece of data that already exists in kube_store
2) Create a patch command to modify that existing data
3) Issue the patch command
4) Check the data in kube_store and verify it was modified as you expected.

src/client/kube_store.test.ts
andyvesel: I've added { 'spec': {patch: {'test': 'foo'}}}

src/client/kube_store/index.ts
andyvesel: fixed

src/client/kube_store/index.ts
andyvesel: fixed

src/client/kube_store/index.ts
andyvesel: If I follow this correctly, the localStorage should contain the entire mocked data + patch, so it'd be like 
```
    const localStorageData = {...mocked_data, ...patchedObject};
    localStorage.setItem(localStorageMockedDataKey, JSON.stringify(localStorageData))
```

src/client/kube_store/index.ts
```diff
@@ -221,6 +35,58 @@ export default class KubeStore {
     return d['clusters'][this.clusterName];
   }
 
+  private determineKey(resource: KubeResource, name: string): object {
```
jwmatthews: Note this is different than what we discussed.

We discussed that we'd refactor out a method to determine the key, hence the name of this function is 'determineKey', yet this function as written now is returning the actual value and not the key.

src/client/kube_store.test.ts
```diff
@@ -36,6 +37,7 @@ const examplePlan = {
 const testNs = 'test-ns';
 const planName = 'my-plan';
 const expectedGvk = 'migration.openshift.io/v1alpha1/migplans';
+const clusterName = 'foo';
```
jwmatthews: Why is this here?

src/client/kube_store/index.ts
```diff
@@ -233,6 +81,20 @@ export default class KubeStore {
     return this.instance || (this.instance = new this(''));
   }
 
+  public patchResource(resource: KubeResource, name: string, patch: object): object {
+    const key = this.determineKey(resource, name);
+    const currentResource = this.data()[key][name];
+    const patchedObject = JsonMergePatch.apply(currentResource, patch);
+    const localStorageData = {...mocked_data, ...patchedObject};
```
jwmatthews: This is not creating the structure as you want.
This is taking 'patchedObject' and putting it at the root level of the object, instead of updating it in the nested key it needs to go in.

jwmatthews:
I will come back to help with this PR once we've gotten past the e2e milestone.

Last I looked we need to fix how patched data is being stored in localStorage.

eriknelson:
Going to close this for now. There's probably some useful work here but it can be referenced at a future time.

eriknelson:
Branch is available in the `patch-request` branch in `fusor/mig-ui`.

Issue #232, opened 2019-06-05
Implement Logout


eriknelson:
Logout is not implemented currently. Should clear the token from local storage and redirect to root.

eriknelson:
The desire for this was to land the user back on the openshift login page, but while implementing this, it became clear that Openshift's IDP will maintain the user session always, until expiry. We don't have any control over this from our perspective. The security team suggested this is not required for GA, and an improved experience would be to learn what URL the standard console is redirecting to during logout, and to piggyback on that. The plan is to implement this post 4.2.

CC: @jwmatthews

Issue #158, opened 2019-05-02
Replace username with simple logout link


ibolton336:

eriknelson:
I'm not actually sure this is even available to us in the oauth response, or if we just need this to be a logout button?

eriknelson:
The username is not available to us, decision is to remove the current dropdown and replace with a logout link.

PR #332, opened 2019-07-12
Add token expiry and logout fn

### LINKED ISSUE #216: Using the 'yarn start:remote' option, after ~2 days, Loading icons are present and not able to load data from k8s API.
After 2-3 days of having 'yarn start:remote' up and working on UI code I've lost access to talk to k8s cluste.r . Assume a token has expired.

At a minimum we need to detect this error and report back to UI, right now I don't see any errors displayed, had to look at js console to find permissions issues.

Seeing 401 and 403 errors.

{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "migclusters.migration.openshift.io is forbidden: User \"system:anonymous\" cannot list resource \"migclusters\" in API group \"migration.openshift.io\" in the namespace \"mig\"",
  "reason": "Forbidden",
  "details": {
    "group": "migration.openshift.io",
    "kind": "migclusters"
  },
  "code": 403
}

Failed to load resource: the server responded with a status of 401 ()
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "Unauthorized",
  "reason": "Unauthorized",
  "code": 401
}

### LINKED ISSUE #228: Implement token expiry and re-login
Currently if the user's token expire's, the UI basically stalls completely and can't talk to anything, because the token is no longer valid. On startup, prior to login, we should confirm that the token is not expired, and if it is, clear the token from localstorage and reload, which will force the user to login again and obtain a new valid token.

Fixes #216

eriknelson:
Closes #228 
Closes #216

Adds the logout function that should really just be called to take care of #232 and #158.

Some of the motivations for this approach: https://github.com/fusor/mig-ui/issues/228#issuecomment-506033628

Notably:
> We don't want every caller to the client to define token expiry behavior; it should be easy to get a client and the client should default to that behavior

> It's desirable to keep the client decoupled from redux. It's likely we'd need to dispatch some router action to move to a new location (/login), but it's questionable the client should be calling redux directly

Think this achieves that.

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -2,5 +2,6 @@
   "baseUrl": "http://localhost:9000",
   "video": false,
   "pluginsFile": false,
-  "supportFile": false
+  "supportFile": false,
+  "chromeWebSecurity": false
 }
--- a/src/app/auth/LoginComponent.tsx
+++ b/src/app/auth/LoginComponent.tsx
@@ -28,7 +28,11 @@ class LoginComponent extends React.Component<IProps> {
     const routerLoc = this.props.router.location;
 
     const freshOauthMeta = !prevProps.auth.oauthMeta && !!oauthMeta;
-    if (freshOauthMeta) {
+    const urlParams = new URLSearchParams(window.location.search);
+    const shouldRefresh = urlParams.get('action') === 'refresh';
+    const handleLogin = freshOauthMeta || shouldRefresh;
+
+    if (handleLogin) {
       const clusterAuth = new ClientOAuth2({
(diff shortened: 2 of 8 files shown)


eriknelson:
I saw this break compass tests for inexplicable reasons, so if they still show up in the travis log I'll have to dig into it. I don't believe the errors had anything to do with the PR, so we may have to repair compass or disable it until we have the bandwidth to fix.

Issue #318, opened 2019-07-10
Remove flash message of 'Found PVs'


jwmatthews:
Currently in MigPlan Wizard, we are flashing a message of Found PVs when we enter PV Discovery.

I think we should remove that, seeing PVs is expected, feels like flash message detracts from UXD

eriknelson:
Fixed by #340

PR #339, opened 2019-07-16
Wizard results and status polling/ pv poll refactor

### LINKED ISSUE #293: Need to implement validation logic step at end of MigPlan wizard
I'm thinking this is a matter of watching the plan for a ready condition once everything has been PUT to the object at that step.

ibolton336:
-- extracted extra functions to utils file
-- changed functions and params dealing with polling to more declarative names
-- expanded checkMigrationStatus poll /status polling saga to check plan status also
Closes #293

### DIFF
--- a/src/app/plan/components/Wizard/ResultsStep.tsx
+++ b/src/app/plan/components/Wizard/ResultsStep.tsx
@@ -6,75 +6,39 @@ import Loader from 'react-loader-spinner';
 import styled from '@emotion/styled';
 import theme from '../../../../theme';
 import { Flex, Box, Text } from '@rebass/emotion';
+import PlanStatus from '../../../home/components/DataList/Plans/PlanStatus';
 import StatusIcon from '../../../common/components/StatusIcon';
 const StyledSpan = styled.span`
   font-weight: 600;
 `;
 
-interface IState {
-  page: number;
-  perPage: number;
-  pageOfItems: any[];
-  rows: any;
-  selectAll: any;
-  checked: any;
-}
-interface IProps {
-  values: any;
-}
+const ResultsStep = props => {
+  const { values, errors, planList } = props;
+  const matchingPlan = planList
+    .filter(p => {
+      return values.planName === p.MigPlan.metadata.name;
+    })
+    .pop();
+  return (
+    <Flex
+      css={css`
+        height: 100%;
+        text-align: center;
+      `}
+    >
+      <Box flex="1" m="auto">
+        <Text fontSize={[2, 3, 4]}>
+          {matchingPlan.PlanStatus.hasReadyCondition ? (
+            <StatusIcon isReady={true} />
+          ) : (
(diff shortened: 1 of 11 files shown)


src/app/plan/duck/operations.ts
```diff
@@ -375,6 +285,74 @@ function* fetchPlansGenerator() {
...
+const getPVs = (dispatch, updatedPlansPollingResponse, newObjectRes) => {
+  const matchingPlan = updatedPlansPollingResponse.updatedPlans
+    .filter(p => p.MigPlan.metadata.name === newObjectRes.data.metadata.name)
+    .pop();
+
+  const pvSearchStatus = matchingPlan ? getPlanPVs(matchingPlan) : null;
+  if (pvSearchStatus.success) {
+    dispatch(Creators.updatePlan(matchingPlan.MigPlan));
+    dispatch(pvFetchSuccess());
+    dispatch(commonOperations.alertSuccessTimeout('Found PVs!'));
```
eriknelson: We have an issue to ditch this notification, could knock out another one with this PR: https://github.com/fusor/mig-ui/issues/318

src/app/plan/duck/operations.ts
```diff
@@ -19,6 +17,9 @@ import {
 } from '../../../client/resources/conversions';
 import { commonOperations } from '../../common/duck';
 import { isSelfSignedCertError, handleSelfSignedCertError } from '../../common/duck/utils';
+import { groupPlan, groupPlans, getMigrationStatus, getPlanStatus, getPlanPVs } from './utils';
```
eriknelson: I almost want to import this as `import planUtils from './utils'` so the usage reads: `planUtils.getPlanPVs`. Makes things more clear where they're coming from. What do you think?

src/app/plan/components/Wizard/ResultsStep.tsx
```diff
@@ -6,75 +6,39 @@ import Loader from 'react-loader-spinner';
...
-  rows: any;
-  selectAll: any;
-  checked: any;
-}
-interface IProps {
-  values: any;
-}
+const ResultsStep = props => {
+  const { values, errors, planList } = props;
+  const matchingPlan = planList
```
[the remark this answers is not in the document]
eriknelson: bunch of other places this could be used instead of a `filter(...).pop()`, if the latter only ever expects to find one element.

eriknelson:
Couple additional comments:
* Tried to add a plan, saw the background polling and it finally looks like it hit some stop condition. I don't know if that's a failure or a success, but I did not see the validation screen update; it's stuck on:
image
* I think that screen needs some improvement from a UX perspective. You arrive there and it just looks like it's error'd and static. I think it needs to suggest we're watching for a ready condition with some kind of a spinner, and the red exclamation mark is harsh.
* The polling interval seems way too fast. It should probably be cut in half.

PR #340, opened 2019-07-16
Status polling wizard results

ibolton336:


ibolton336:
> This is a really big PR, so I've done my best to review it. I added requested changes to a previous PR that have been addressed in this one. I've also run through the migration scenarios without issue, and saw the added improvements occur (clusters becoming ready as expected, over time).
> 
> My request would be: we have to split these PRs into smaller ones like: refactor existing logic to be reusable, then cluster watch, then storage watch, then plan validation. It's impossible to review something this large that conflates this many issues.

10-4 I will definitely keep this in mind going forward.

Issue #323, opened 2019-07-11
Quiesce should be false (at least by default) when stage is true


jmontleon:
I think this is being set here

https://github.com/fusor/mig-ui/blob/42e09acf2464b70206e35e96a0aaa7f77ad5d02b/src/client/resources/conversions.ts#L291

eriknelson:
To be clear on this one, quiesce pods should be false on stages, but should be controlled by the checkbox in the confirmation modal when creating a final migration.

Issue #224, opened 2019-06-05
Need to implement quiescePods checkbox to set the right field on the MigMigration


eriknelson:
https://github.com/fusor/mig-ui/pull/218#discussion_r290807746

Value is hardcoded on the MigMigration currently

eriknelson:
Closed by #351

PR #341, opened 2019-07-16
Fixed status propagaton on failed migrations

### LINKED ISSUE #315: mig-ui says migration succeeded when it failed
migmigration

```
Name:         72188780-a195-11e9-95aa-f318654dbe30
Namespace:    mig
Labels:       <none>
Annotations:  touch=1657d18f-736a-4c6c-834c-bf37d0d3119f
API Version:  migration.openshift.io/v1alpha1
Kind:         MigMigration
Metadata:
  Creation Timestamp:  2019-07-08T15:31:24Z
  Generation:          4
  Resource Version:    3930797
  Self Link:           /apis/migration.openshift.io/v1alpha1/namespaces/mig/migmigrations/72188780-a195-11e9-95aa-f318654dbe30
  UID:                 722bc9cd-a195-11e9-a219-0ed41539d98e
Spec:
  Mig Plan Ref:
    Name:        test
    Namespace:   mig
  Quiesce Pods:  true
  Stage:         false
Status:
  Conditions:
    Category:              Advisory
    Last Transition Time:  2019-07-08T15:31:37Z
    Message:               The migration has completed successfully.
    Reason:                BackupFailed
    Status:                True
    Type:                  Succeeded
    Category:              Critical
    Last Transition Time:  2019-07-08T15:31:37Z
    Message:               The migration has failed.  See: Errors.
    Reason:                BackupFailed
    Status:                True
    Type:                  Failed
  Errors:
    Backup: mig/72188780-a195-11e9-95aa-f318654dbe30-68fk6 failed.
  Phase:            BackupFailed
  Start Timestamp:  2019-07-08T15:31:24Z
Events:             <none>
```

ui displays:
```
test 1 origin3-dev host jmontleo-velero 0 Migration Succeeded  
```

The icon on the left is red instead of green like the one that succeeded so it seems the actual state is detected and maybe just the text needs fixed for this case.

Danil-Grigorev:
Here is a case, where the fix makes a change:
Before - last status gets propagated to the failed builds, which has no migrations executed:
image
Appearance after the fix:
image (1)

### DIFF
--- a/src/app/home/components/DataList/Plans/PlanDataListItem.tsx
+++ b/src/app/home/components/DataList/Plans/PlanDataListItem.tsx
@@ -39,7 +39,7 @@ const PlanDataListItem = ({
               toggleExpanded();
             }}
             isExpanded={isExpanded}
-            id="cluster-toggle"
+            id="plan-toggle"
           />
           <DataListItemCells
             dataListCells={[
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -14,18 +14,17 @@ const getAllPlans = createSelector(
 const getPlansWithStatus = createSelector(
   [planSelector],
   plans => {
-    let hasReadyCondition = null;
-    let hasPlanError = null;
-    let hasMigrationError = null;
-    let hasPrevMigrations = null;
-    let hasRunningMigrations = null;
-    let finalMigrationComplete = null;
-    let hasSucceededMigration = null;
-    let hasSucceededStage = null;
-    let hasClosedCondition = null;
-    let latestType = null;
-
     const plansWithStatus = plans.map(plan => {
+      let hasReadyCondition = null;
+      let hasPlanError = null;
+      let hasMigrationError = null;
+      let hasPrevMigrations = null;
+      let hasRunningMigrations = null;
+      let finalMigrationComplete = null;
+      let hasSucceededMigration = null;
+      let hasSucceededStage = null;
+      let hasClosedCondition = null;
+      let latestType = null;
       if (plan.MigPlan.status) {
         hasClosedCondition = plan.MigPlan.spec.closed;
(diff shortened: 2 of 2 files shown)


src/client/resources/conversions.ts
```diff
@@ -282,7 +282,7 @@ export function createMigMigration(
       // TODO: We need to wire these up to the UI when stage is implemented
       // and we've more thouroughly tested quiesce/stage combos in conjunction
       // with the controller and velero.
-      quiescePods: true,
+      quiescePods: !isStage,
```
eriknelson: This is tricky, and it's not strictly a function of whether or not we're doing a stage. During a migration, the user is presented an option to quiesce or not. I would remove this from the PR and hit later in a distinct PR, since it's more nuanced. See: https://github.com/fusor/mig-ui/issues/224

src/app/plan/duck/operations.ts
```diff
@@ -241,7 +241,6 @@ const addPlan = migPlan => {
           if (pvsDiscovered) {
             dispatch(Creators.updatePlan(plan));
             dispatch(pvFetchSuccess());
-            dispatch(commonOperations.alertSuccessTimeout('Found PVs!'));
```
eriknelson: This has been addressed in #340, it's fine to leave but just letting you know.

eriknelson:
You'll have to rebase on the recent PR we added.

src/client/resources/conversions.ts
Danil-Grigorev: Ok, I'll try to implement check box there later. Thanks for the suggestion, the issue https://github.com/fusor/mig-ui/issues/323 implied that it should be connected directly to the `stage` logic for now.

eriknelson:
This is definitely fixing a bug in the status logic, but I don't think it's actually fixing #315. That issue describes the status reporting success when a migration has **failed**, but in your screenshots, you haven't actually run any migrations on your other plans. As far as I'm aware, we don't actually have an issue for what you're demonstrating, so you might have found and fixed a new bug.

PR #349, opened 2019-07-17
Fix hung pv discovery ensure it times out

### LINKED ISSUE #348: PV discovery hangs, and makes the page freeze
During last testing encountered this behaviour. Happens when the creation of `mig-plan` is started from the UI. Independent on the selected namespace and presence of the PVs in the namespace, the discovery runs for some time and then blocks the window indefinitely.

eriknelson:
Closes #348 

If pv discovery ever "timed out" (ran past the desired amount of ticks), it would fall into an infinite while loop that would deadlock the UI thread. This at least will prevent a deadlock and alert the user to a timeout.

### DIFF
--- a/src/app/plan/duck/sagas.ts
+++ b/src/app/plan/duck/sagas.ts
@@ -1,7 +1,7 @@
 import { race, call, delay, put, take } from 'redux-saga/effects';
 
 import { Creators } from './actions';
-import { alertError } from '../../common/duck/actions'
+import { alertError } from '../../common/duck/actions';
 
 const TicksUntilTimeout = 20;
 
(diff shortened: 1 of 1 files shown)


src/app/plan/duck/sagas.ts
```diff
@@ -26,6 +29,14 @@ function* checkPVs(action) {
           break;
       }
       yield delay(params.delay);
+    } else {
+      // PV discovery timed out, alert and stop polling
+      pvsFound = true; // No PVs timed out
+      Creators.stopPVPolling();
+      yield put(alertError("Timed out during PV discovery"));
```
ibolton336: Looks good & nice catch 👍

Issue #346, opened 2019-07-17
When a migplan is created and in a not ready state, upper right card thinks its complete


eriknelson:
![image](https://user-images.githubusercontent.com/410933/61386221-e2278880-a881-11e9-860b-4fb1d869d082.png)

I created a migplan referring to a source cluster, then I deleted that source cluster. That will throw the plan into a "Not Ready" state with the following critical condition:

```
    conditions:
    - category: Critical
      lastTransitionTime: 2019-07-17T14:50:54Z
      message: The `srcMigClusterRef` must reference a `migcluster`.
      reason: NotFound
      status: "True"
      type: InvalidSourceClusterRef
```

This causes the UI to display an incorrect state in the upper right part of the card as seen in the screenshot.

PR #350, opened 2019-07-17
Fixed coupling errored plans with completed

Danil-Grigorev:
Closes https://github.com/fusor/mig-ui/issues/346

Getting from:
image
To:
image
(One plan was added later during other testing)

### DIFF
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -103,9 +103,6 @@ const getCounts = createSelector(
         hasErrorCondition = !!plan.MigPlan.status.conditions.filter(c => c.category === 'Critical')
           .length;
 
-        if (hasErrorCondition) {
-          counts.completed.push(plan);
-        }
         if (plan.Migrations.length) {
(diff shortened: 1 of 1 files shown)


eriknelson:
@Danil-Grigorev I put a do not merge on this because I think we want to have a broader discussion about exactly what "completed" means; I think it's more nuanced than this conditional, but I'm not sure yet exactly what our desired behavior is. Ex:

* I have a new migplan that has **never** been staged or migrated, and has never been ready. It doesn't make sense to me to consider this as "Completed". It hasn't completed anything, since it's been broken.
* I have a migplan that has run a migration/stage, and that has failed for some reason. It's fair to consider this "completed" in an error state, since it actually did some work.

Bookmarking this for some discussion when I can get ahold of other folks to discuss.

eriknelson:
This correctly addresses the first case. As far as the second case is concerned, we have more work to do in the status area that's in the pipeline, so it's likely to be resolved by that work and is outside the scope of this PR.

PR #351, opened 2019-07-18
Added quiesce pod logic handling from the migrate modal

Danil-Grigorev:
Closes https://github.com/fusor/mig-ui/issues/323 and https://github.com/fusor/mig-ui/issues/224


eriknelson:
Tested all these scenarios and looks good :+1:

PR #363, opened 2019-07-23
Migstorage secret discovery refactor

### LINKED ISSUE #362: If a MigStorage fails to load, all of the MigStorages fail
Turns out that if user have predefined `migstorages` with different relation between secret and the resource itself (naming does not match), the UI looses record of such resources.

Danil-Grigorev:
- Improved resiliency.
- Driven by ref name.
Closes #362

### DIFF
--- a/src/app/home/HomeComponent.tsx
+++ b/src/app/home/HomeComponent.tsx
@@ -144,12 +144,10 @@ class HomeComponent extends React.Component<IProps, IState> {
   };
 
   handleStoragePoll = response => {
-    if (response && response.isSuccessful === true) {
+    if (response) {
       this.props.updateStorages(response.updatedStorages);
-      return true;
     }
-
-    return false;
+    return response.isSuccessful;
   };
 
   componentDidMount = () => {
--- a/src/app/home/components/Card/Status/CardStatus.tsx
+++ b/src/app/home/components/Card/Status/CardStatus.tsx
@@ -6,16 +6,23 @@ import { Grid, GridItem } from '@patternfly/react-core';
 
 const CardStatusComponent = ({ type, dataList, ...props }) => {
   let successList = [];
+  let failureList = [];
+  let successfulNames = [];
+
   if (type === 'repositories') {
     successList = dataList.filter(item =>
       item.MigStorage.status && item.MigStorage.status.conditions.filter(c => c.type === 'Ready').length !== 0
       );
+    successfulNames = successList.map(item => item.MigStorage.metadata.name);
+    failureList = dataList.filter(item => !successfulNames.includes(item.MigStorage.metadata.name));
   } else if (type === 'clusters') {
     successList = dataList.filter(item =>
       item.MigCluster.status && item.MigCluster.status.conditions.filter(c => c.type === 'Ready').length !== 0
     );
+    successfulNames = successList.map(item => item.MigCluster.metadata.name);
+    failureList = dataList.filter(item => !successfulNames.includes(item.MigCluster.metadata.name));
   }
-  const failureList = dataList.filter(item => successList.indexOf(item) < 0);
+
   return (
     <React.Fragment>
       <Grid>
--- a/src/app/home/components/Card/Status/CardStatus.tsx
+++ b/src/app/home/components/Card/Status/CardStatus.tsx
@@ -6,28 +6,16 @@ import { Grid, GridItem } from '@patternfly/react-core';
 
 const CardStatusComponent = ({ type, dataList, ...props }) => {
   let successList = [];
-  const failureList = [];
   if (type === 'repositories') {
-    // hasReadyCondition = plan.MigPlan.status.conditions.filter(c => c.type === 'Ready').length > 0;
-    // hasErrorCondition =
-    //   plan.MigPlan.status.conditions.filter(c => c.category === 'Critical').length > 0;
-
-    successList = dataList.filter(item => {
-      if (item.MigStorage.status) {
-        return item.MigStorage.status.conditions.filter(c => c.type === 'Ready');
-      } else {
-        failureList.push(item);
-      }
-    });
+    successList = dataList.filter(item =>
+      item.MigStorage.status && item.MigStorage.status.conditions.filter(c => c.type === 'Ready').length !== 0
+      );
   } else if (type === 'clusters') {
-    successList = dataList.filter(item => {
-      if (item.MigCluster.status) {
-        return item.MigCluster.status.conditions.filter(c => c.type === 'Ready');
-      } else {
-        failureList.push(item);
-      }
-    });
+    successList = dataList.filter(item =>
+      item.MigCluster.status && item.MigCluster.status.conditions.filter(c => c.type === 'Ready').length !== 0
+    );
   }
+  const failureList = dataList.filter(item => successList.indexOf(item) < 0);
   return (
     <React.Fragment>
       <Grid>
--- a/src/app/storage/duck/operations.ts
+++ b/src/app/storage/duck/operations.ts
@@ -157,8 +157,8 @@ const fetchStorage = () => {
       const res = await client.list(resource);
       //temporary for ui work
       const migStorages = res.data.items;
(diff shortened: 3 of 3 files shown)


src/app/storage/duck/operations.ts
```diff
@@ -159,8 +159,8 @@ const fetchStorage = () => {
       const res = await client.list(resource);
       //temporary for ui work
       const migStorages = res.data.items;
-      const refs = await Promise.all(fetchMigStorageRefs(client, migMeta, migStorages));
-      const groupedStorages = groupStorages(migStorages, refs);
+      const secretRefs = await fetchMigStorageSecretRefs(client, migMeta, migStorages);
```
eriknelson: +1 with these renames, I was going to suggest we should be more specific with these names as secrets.

This is breaking the way that the flow of the function works. I think we want to use `Promise.allSettled` for this [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled), [Proposal compliant shim](https://www.npmjs.com/package/promise.allsettled), ex:

`const secretRefResults = await Promise.allSettled(fetchMigStorageSecretRefs(client, migMeta, migStorages));`

Then you can determine the set of successful, fully loaded `MigStorages` from the ones that failed.

Following that, you can then dispatch an update and potentially handle failed `MigStorages`, which we should mark as not ready.

src/app/storage/duck/operations.ts
```diff
@@ -185,38 +186,36 @@ function fetchMigStorageRefs(client: IClusterClient, migMeta, migStorages): Arra
...
-    storageList = yield storageList.data.items;
-    const refs = yield Promise.all(fetchMigStorageRefs(client, state.migMeta, storageList));
-    const groupedStorages = groupStorages(storageList, refs);
-    return { updatedStorages: groupedStorages, isSuccessful: true };
-  } catch (e) {
-    return { e, isSuccessful: false };
-  }
+  let storageList = yield client.list(resource);
+  storageList = yield storageList.data.items;
+  const secretRefs = yield Promise.all(fetchMigStorageSecretRefs(client, state.migMeta, storageList));
```
eriknelson: This logic is going to cause problems for the same reasons as above. It should use similar logic to the above function. In fact, the logic I described is likely going to be so close that it should be brought out into a shared function so as not to duplicate the code.

src/app/home/components/Card/Status/CardStatus.tsx
```diff
@@ -6,28 +6,16 @@ import { Grid, GridItem } from '@patternfly/react-core';
...
-        return item.MigCluster.status.conditions.filter(c => c.type === 'Ready');
-      } else {
-        failureList.push(item);
-      }
-    });
+    successList = dataList.filter(item =>
+      item.MigCluster.status && item.MigCluster.status.conditions.filter(c => c.type === 'Ready').length !== 0
+    );
   }
+  const failureList = dataList.filter(item => successList.indexOf(item) < 0);
```
eriknelson: Although this happens to work due to some specific circumstances, it's can be a very error prone way to check objects for semantic equality (`indexOf` uses strict comparison, `===`). It will in fact return (this example does). You have to be extremely careful when you are trying to see if two objects are "equal" to one another, because "equal" can mean very different things depending on the context:

```
const foo = {name: 'hello', deeperThing: { age: 10 }};
const bar = {name: 'hello', deeperThing: { age: 10 }};

console.log(foo === bar)
===
Output: false
```

A better check here would be to retrieve the set of objects that are **not** found in the successful set, based on their names:

```diff
diff --git a/src/app/home/components/Card/Status/CardStatus.tsx b/src/app/home/components/Card/Status/CardStatus.tsx
index 4e7e350..f2cc7d9 100644
--- a/src/app/home/components/Card/Status/CardStatus.tsx
+++ b/src/app/home/components/Card/Status/CardStatus.tsx
@@ -6,16 +6,22 @@ import { Grid, GridItem } from '@patternfly/react-core';
 
 const CardStatusComponent = ({ type, dataList, ...props }) => {
   let successList = [];
+  let failureList = [];
+
   if (type === 'repositories') {
     successList = dataList.filter(item =>
       item.MigStorage.status && item.MigStorage.status.conditions.filter(c => c.type === 'Ready').length !== 0
       );
+    const successfulNames = successList.map(item => item.MigStorage.metadata.name);
+    failureList = dataList.filter(item => !successfulNames.includes(item.MigStorage.metadata.name));
   } else if (type === 'clusters') {
     successList = dataList.filter(item =>
       item.MigCluster.status && item.MigCluster.status.conditions.filter(c => c.type === 'Ready').length !== 0
     );
+    const successfulNames = successList.map(item => item.MigCluster.metadata.name);
+    failureList = dataList.filter(item => !successfulNames.includes(item.MigCluster.metadata.name));
   }
-  const failureList = dataList.filter(item => successList.indexOf(item) < 0);
+
   return (
     <React.Fragment>
       <Grid>
```

src/app/home/HomeComponent.tsx
```diff
@@ -144,12 +144,10 @@ class HomeComponent extends React.Component<IProps, IState> {
   };
 
   handleStoragePoll = response => {
-    if (response && response.isSuccessful === true) {
```
ibolton336: Should move these changes to a different PR targeted at refactoring the polling

src/app/home/components/Card/Status/CardStatus.tsx
```diff
@@ -6,28 +6,23 @@ import { Grid, GridItem } from '@patternfly/react-core';
...
-        return item.MigStorage.status.conditions.filter(c => c.type === 'Ready');
-      } else {
-        failureList.push(item);
-      }
-    });
+  if (type === 'repositories') {
+    successList = dataList.filter(item =>
+      item.MigStorage.status && item.MigStorage.status.conditions.filter(c => c.type === 'Ready').length !== 0
+      );
+    successfulNames = successList.map(item => item.MigStorage.metadata.name);
```
ibolton336: We should move all of this logic to a selector. You are right about needing to refactor this, but while we are fixing it, we should move it out of the presentation layer & into a selector: https://github.com/fusor/mig-ui/blob/master/src/app/plan/duck/selectors.ts

src/app/storage/duck/operations.ts
```diff
@@ -204,15 +214,10 @@ function* fetchStorageGenerator() {
   const state = yield select();
   const client: IClusterClient = ClientFactory.hostCluster(state);
   const resource = new MigResource(MigResourceKind.MigStorage, state.migMeta.namespace);
-  try {
-    let storageList = yield client.list(resource);
-    storageList = yield storageList.data.items;
-    const refs = yield Promise.all(fetchMigStorageRefs(client, state.migMeta, storageList));
-    const groupedStorages = groupStorages(storageList, refs);
-    return { updatedStorages: groupedStorages, isSuccessful: true };
-  } catch (e) {
```
ibolton336: Within generator functions we need to handle the error case by using throw err in a catch block to stop the generator.

src/app/storage/duck/operations.ts
```diff
@@ -181,20 +182,29 @@ function fetchMigStorageRefs(client: IClusterClient, migMeta, migStorages): Arra
       CoreNamespacedResourceKind.Secret,
       secretRef.namespace
     );
-    refs.push(client.get(secretResource, secretRef.name));
+    const resp = client.get(secretResource, secretRef.name);
+    resp.then(r => {
```
ibolton336: we are hoping to standardize across the app using async/await or yield/generators & trying to avoid using legacy promise code.

eriknelson: I'm thinking we probably want to deprecate thunks and legacy promise code unless for some reason they're absolutely necessary.

eriknelson:
We chatted this one over offline and feel that to implement the wider feature here is out of scope for this PR. Submitting a PR that correctly dereferences the secret ref will fix the immediate linked issue (https://github.com/fusor/mig-ui/pull/363/files#diff-d63dd651509698d746484d8ad4394199R203), and we're going to close this one.

PR #364, opened 2019-07-23
Storage class selection implementation

### LINKED ISSUE #259: Implement storageClass assignment for PVs in target cluster
After the PVs have been discovered and the user has designated them either "copy" or "move", the user will need the ability to choose a storage class that will be used in the target cluster for the PV. This would support use cases of migrating NFS backed PVs to something like EBS storage.

The PV discovery process should take care of adding the source PV's storageclass to the discovered PV section, if it exists. (Not present in the controller yet, issue filed: https://github.com/fusor/mig-controller/issues/134). To get the available storageclasses the user can choose from on the target cluster, the UI needs to query that cluster with `LIST storageclasses`.

Here's an example storage class from a target OCP4 cluster, for EBS:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
  creationTimestamp: null
  name: gp2
  ownerReferences:
  - apiVersion: v1
    kind: clusteroperator
    name: storage
    uid: d4bc5ea4-782f-11e9-a701-0a4e52dea9a2
parameters:
  encrypted: "true"
  type: gp2
provisioner: kubernetes.io/aws-ebs
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
```

We also think it would be useful to include the `provisioner` field in the dropdown string, because it adds more information than just a name like `gp2`. Ex dropdown string: `gp2:kubernetes.io/aws-ebs`.

Note: PVs that have been designated "move" should be left out of this list, because the storageclass assignment is not applicable.

FYI @vconzola

### LINKED ISSUE #353: Next should be disabled during PV discovery until it is complete.
With latest updates the condition here is failing https://github.com/fusor/mig-ui/blob/16fefc9f4946694836e52b7209c2f0781af38455/src/app/plan/components/Wizard/WizardComponent.tsx#L83 The UI doesn't start the PV discovery at once (takes 1s) and you can click on Next right away. The picture below describes possible outcome - skipped step.

image (3)

ibolton336:
Closes #353 
Closes #259


eriknelson:
Putting a do not merge on this since we don't want this in master until the controller support has been added. https://github.com/fusor/mig-controller/pull/187

sseago:
@eriknelson The associated controller PR has now been merged.

eriknelson:
Thanks @sseago!

src/app/plan/components/Wizard/VolumesTable.tsx
```diff
@@ -15,80 +15,56 @@ import Loader from 'react-loader-spinner';
...
-
           // TODO: A number of these values will need to be further supported by the
           // controller. Today the data is not available.
           // See the mig controller issue describing what data we need here:
           // https://github.com/fusor/mig-controller/issues/134
           return {
             name: planVolume.name,
             project: '',
-            storageClass: '',
+            storageClass: planVolume.selection.storageClass || '',
```
ibolton336: @eriknelson when discovering PVs, there is already a default selection on the PV for storage class taht is added to values.persistentVolumes here. On the next step in the storageClass selection, you can change the object value on the PV to the selection of your choice in the dropdown. see: https://github.com/fusor/mig-ui/pull/364/files#diff-2848a72f04be20e44dd3940a1f78c2b9R29

ibolton336: Still need to figure out a way to test multiple storage classes for selection, though.

src/client/resources/conversions.ts
```diff
@@ -217,10 +217,11 @@ export function updateMigPlanFromValues(migPlan: any, planValues: any) {
     namespace: migPlan.metadata.namespace,
   };
 
-  if(updatedSpec.persistentVolumes) {
+  if (updatedSpec.persistentVolumes) {
     updatedSpec.persistentVolumes = updatedSpec.persistentVolumes.map(v => {
       const userPv = planValues.persistentVolumes.find(upv => upv.name === v.name);
-      v.action = userPv.type;
+      v.selection.action = userPv.type;
+      v.selection.storageClass = userPv.storageClass;
```
ibolton336: storageClass is updated on the migPlan when the updatePlan is run at the end of the wizard.

src/app/plan/components/Wizard/StorageClassTable.tsx
```diff
@@ -0,0 +1,152 @@
...
+import 'react-table/react-table.css';
+import Select from 'react-select';
+import { css } from '@emotion/core';
+import { Flex, Box, Text } from '@rebass/emotion';
+import theme from '../../../../theme';
+import Loader from 'react-loader-spinner';
+
+const StorageClassTable = (props): any => {
+  const { planList, clusterList, values, isFetchingPVList } = props;
+  const [rows, setRows] = useState([]);
```
eriknelson: Man this really feels a lot cleaner than previous stuff. :+1:

src/app/plan/duck/operations.ts
```diff
@@ -41,7 +42,7 @@ const namespaceFetchFailure = Creators.namespaceFetchFailure;
 const updatePlanResults = Creators.updatePlanResults;
 const updatePlan = Creators.updatePlan;
 
-const PlanMigrationPollingInterval = 1000;
+const PlanMigrationPollingInterval = 5000;
```
eriknelson: :+1:

PR #367, opened 2019-07-23
S3 bucket select

### LINKED ISSUE #360: s3 bucket region can be a select input
There is a limited amount of regions so having this field as select input can be a better UX.
Screenshot 2019-07-22 at 12 31 36

Danil-Grigorev:
Closes #360, in case of a change approval.


eriknelson:
Going to close this for now so we don't have draft PRs sitting open for weeks that are likely to fall significantly behind, or require larger changes to the point where this is obsolete. The issue this is addressing is not a beta blocker, so it's unlikely it will be looked at for some time.

PR #368, opened 2019-07-23
retry function for put plan

### LINKED ISSUE #222: MigPlan update needs retry logic in case the response 409s
In the GET, mutate, PUT workflow, it's possible the resourceVersion has changed out from under that operation (usually by the controller). If the resourceVersion in our put payload doesn't match what's on the server, it will reject it with a 409. We need retry logic around this in `putPlan`.

ibolton336:
Closes #222

### DIFF
--- a/src/app/plan/duck/operations.ts
+++ b/src/app/plan/duck/operations.ts
@@ -230,6 +230,7 @@ const addPlan = migPlan => {
   };
 };
 
+<<<<<<< HEAD
 const putPlan = planValues => {
   return async (dispatch, getState) => {
     try {
@@ -261,6 +262,8 @@ const putPlan = planValues => {
   };
 };
 
+=======
+>>>>>>> putPlan cleanup
 const removePlan = id => {
   throw new Error('NOT IMPLEMENTED');
 };
@@ -352,7 +355,6 @@ export default {
   pvFetchRequest,
   fetchPlans,
   addPlan,
-  putPlan,
   removePlan,
   fetchNamespacesForCluster,
   runStage,
--- a/src/app/plan/duck/operations.ts
+++ b/src/app/plan/duck/operations.ts
@@ -230,40 +230,6 @@ const addPlan = migPlan => {
   };
 };
 
-<<<<<<< HEAD
-const putPlan = planValues => {
-  return async (dispatch, getState) => {
-    try {
-      const state = getState();
-      const migMeta = state.migMeta;
-      const client: IClusterClient = ClientFactory.hostCluster(state);
-
-      // When updating objects
-      const latestPlanRes = await client.get(
-        new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
-        planValues.planName
-      );
-      const latestPlan = latestPlanRes.data;
-
-      dispatch(Creators.updatePlan(latestPlan));
-      const updatedMigPlan = updateMigPlanFromValues(latestPlan, planValues);
-
-      const putRes = await client.put(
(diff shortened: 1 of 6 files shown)


src/app/plan/duck/operations.ts
```diff
@@ -242,14 +242,23 @@ const putPlan = planValues => {
...
 
-      const putRes = await client.put(
-        new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
-        latestPlan.metadata.name,
-        updatedMigPlan
-      );
-      // TODO: Need some kind of retry logic here in case the resourceVersion
-      // gets ticked up in between us getting and putting the mutated object back
-      dispatch(Creators.updatePlan(putRes.data));
+      const fetch_retry: any = async n => {
```
eriknelson: Can you update the var names here to be more descriptive, something like `i -> currentRetry` and `n -> totalRetries`. you could also initialize `i` here to be 1 and it would incrementing it when checking `isLastAttempt`.

ibolton336:
> Trying to actually test this and i'm not sure how, have you figured out a reproducer?

Try cranking up the plans poll delay

src/app/plan/duck/operations.ts
```diff
@@ -242,14 +242,23 @@ const putPlan = planValues => {
...
+            );
+          } catch (err) {
+            const isLastAttempt = i + 1 === n;
+            if (isLastAttempt) {
+              throw err;
+            }
+          }
+        }
+      };
+      fetch_retry(3).then(res => dispatch(Creators.updatePlan(res.data)));
```
eriknelson: This promise without a `catch` actually results in an unhandled exception despite the code being inside of a try catch. I think you might want to turn this into a regular `await`:
```
const res = await fetchRetry(3);
dispatch(Creators.updatePlan(res.data));
```

If that thing throws, it should fall into the `catch` below as you'd expect.

src/app/plan/duck/operations.ts
eriknelson: NIT: Can you update the var names here to be more descriptive, something like `i -> currentTry` and `n -> totalTries`. you could also initialize i here to be 1 and avoid incrementing it when checking isLastAttempt (`currentTry <= totalTries` too in the loop condition). Also, the rest of our local vars are camel cased, so maybe `fetchRetry` should be too.

src/app/plan/duck/operations.ts
ibolton336: very nice - will change it

src/app/plan/duck/operations.ts
```diff
@@ -242,23 +242,25 @@ const putPlan = planValues => {
...
+        for (let currentTries = 0; currentTries < totalTries; currentTries++) {
           try {
             return await client.put(
               new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
               latestPlan.metadata.name,
               updatedMigPlan
             );
           } catch (err) {
-            const isLastAttempt = i + 1 === n;
+            dispatch(commonOperations.alertErrorTimeout('Failed to update plan. Retrying...'));
```
eriknelson: I'm not sure we want to spam them on retries, could be a relatively normal event and this might make it seems like something is badly wrong.

src/app/plan/duck/operations.ts
```diff
@@ -230,37 +230,6 @@ const addPlan = migPlan => {
   };
 };
 
-const putPlan = planValues => {
```
eriknelson: :+1:

eriknelson:
Tested this and faked a 400, looks good :+1:

PR #377, opened 2019-07-25
Add popover link with pv details

### LINKED ISSUE #312: Implement 'View Details' link in PV Discovery
When building a Plan and selecting PVs, the 'View Details' link for each PV has not been implemented.

ibolton336:
Closes #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,13 +58,8 @@
     "webpack-dev-server": "^3.2.1"
   },
   "dependencies": {
-    "uuid": "^3.3.2",
-    "reduxsauce": "^1.0.1",
-    "react-select": "^2.4.2",
-    "react-loader-spinner": "^2.3.0",
-    "formik": "^1.5.2",
-    "@emotion/styled": "^10.0.9",
     "@emotion/core": "^10.0.10",
+    "@emotion/styled": "^10.0.9",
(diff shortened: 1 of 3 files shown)


ibolton336:
Sent from my iPhone

> On Jul 26, 2019, at 5:24 AM, Erik Nelson <notifications@github.com> wrote:
> 
> @eriknelson requested changes on this pull request.
> 
> This is showing the limited pv details that we're getting back from the controller, but I think we actually want to query the selected source cluster and display the pv object itself. It would be the output from oc get pv <pvname> -o json on the source cluster. I've always had reservations about this because of how much data is actually there, but maybe we can experiment with the size of it to try to make it look better.
> 
Sounds good - will look into adding a PV query to this & report back with how much of that data will be useful to display.
> Also, the popover is very sensitive. I'd mouse over it and it would randomly disappear.
> 
Yes this is an issue with how often we are re-rendering our components due to background poll. Also seeing drop downs auto close due to the same issue. Just opened an issue for this yesterday.

eriknelson:
Talked this one over with @ibolton336 and we agreed it made sense to close this and focus on solving the optimized renders rather than merging what looks like a broken feature.

PR #381, opened 2019-07-26
Removal of reduxsauce

### LINKED ISSUE #379: Drop reduxsauce for strongly typed action creators
This is probably long overdue; we've had a lot of bugs thanks to reduxsauce and wrong types getting stuffed into action creators. It's great at reducing boilerplate, but its too dynamic to play nice with typescript, and we'll be healthier with action creators that are strongly typed and accept arguments that are type checked at compile time.

I would advocate for the following:
* Action types should be `consts` that are exported on a `Types` object to maintain compatibility. Anyone referring to these action types should use the value on `Types`; not a bare string, ex:

* Most actions are single files, we should split them up into separate files, organized by function, under an `actions/` dir with an `index.ts`. See https://github.com/fusor/mig-ui/pull/376 for an example.

* Also export a `Creators` object with all the creators on it for backwards compat.

Example:
```ts
export const Types = {
  FOO_ACTION: 'FOO_ACTION',
}

const fooActionCreator = () => {
  type: Types.FOO_ACTION
}

export const Creators = {
  fooActionCreator,
}
```

Danil-Grigorev:
Dropping reduxsause dependency.
    
- Added strongly typed actions.
- Merging actions into separate file.
- TODO: add interfaces for migration resources.

Closes #379


src/app/auth/duck/errorActions.ts
```diff
@@ -0,0 +1,13 @@
+const certErrorOccurred = (failedUrl: string) =>
```
eriknelson: Not sure about this split, I don't think there are enough auth actions to warrant this. Also, if you're going to split off actions into their own files, the pattern is to turn `actions` into a directory, ex: `src/app/auth/duck/actions/*` with an `index.ts` file that imports the files that you have split up. That way a consumer can use the import path `import { Creators } from './src/app/auth/duck/actions'`. It will load the index.ts and run that code. You might have to re-export things from there. It makes the directory appear as if it's a single module, and allows us to internally organize it into separate files.

src/app/auth/duck/operations.ts
```diff
@@ -23,13 +20,13 @@ const fetchOauthMeta = clusterApi => {
   return async dispatch => {
     try {
       const res = await axios.get(oauthMetaUrl);
-      dispatch(setOauthMeta(res.data));
+      dispatch(Creators.setOauthMeta(res.data));
```
eriknelson: I'm actually more in favor of this vs. aliasing them higher in the file; it makes the call very explicit as far as where it's coming from when reading the calling site.

src/app/auth/duck/reducers.ts
```diff
@@ -1,5 +1,6 @@
 import { createReducer } from 'reduxsauce';
 import { Types } from './actions';
+import { ErrorTypes } from './errorActions';
```
eriknelson: Another issue with breaking this into its own file and changing the export name is you're breaking compatibility with our existing code. We should seek to maintain API compatibility with this change.

src/app/cluster/duck/change_actions.ts
```diff
@@ -0,0 +1,83 @@
...
+const watchClusterAddEditStatus = (clusterName) => ({
+  type: ChangeTypes.WATCH_CLUSTER_ADD_EDIT_STATUS,
+  clusterName,
+});
+
+const cancelWatchClusterAddEditStatus = () => ({
+  type: ChangeTypes.CANCEL_WATCH_CLUSTER_ADD_EDIT_STATUS,
+});
+
+export const ChangeTypes = {
```
eriknelson: I'd personally prefer the types to be listed at the top of the file, ahead of their usage.

src/app/cluster/duck/operations.ts
```diff
@@ -99,15 +97,6 @@ const removeCluster = name => {
   };
 };
 
-function checkConnection() {
```
eriknelson: +1

src/app/cluster/duck/sagas.ts
```diff
@@ -250,11 +252,11 @@ function* startWatchingClusterAddEditStatus(action) {
   const statusToDispatch = addEditResult || createAddEditStatus(
     AddEditState.TimedOut, AddEditMode.Edit);
 
-  yield put(Creators.setClusterAddEditStatus(statusToDispatch));
+  yield put(ChangeActions.setClusterAddEditStatus(statusToDispatch));
 }
 
 function* watchClusterAddEditStatus() {
-  yield takeLatest(Creators.watchClusterAddEditStatus().type, startWatchingClusterAddEditStatus);
```
eriknelson: +1, I think this is more canonical and a convention we should enforce going forwards.

src/app/plan/duck/actions/change_actions.ts
```diff
@@ -0,0 +1,104 @@
+
+
+interface IMigPlan {
```
eriknelson: Exactly what I was hoping for with this PR, looking to bring some strong typing to the action interfaces. :tada:

src/app/auth/duck/index.ts
```diff
@@ -1,4 +1,4 @@
 import authReducer from './reducers';
 export { default as authOperations } from './operations';
-export { Creators as authActionCreators } from './actions';
+export { AuthActions as authActionCreators } from './actions';
```
eriknelson: Don't think this requires a change, but you could just export this as `Actions`, because the import will likely look like:

```
import authDuck from './auth/duck'
...
authDuck.Actions.Creators.callFoo()
```

What you have works too though, because we also are doing a lot of importing the actions directly:

```
import { Actions as AuthActions } from './auth/duck/actions'
```
Your solution here lets us skip that import alias.

Issue #257, opened 2019-06-12
Add cluster/repo/plan modals should reflect the action being performed


eriknelson:
Currently the titles are simply "Cluster" or "Storage", but as @vconzola specified, the title of the modals should reflect the action that's being performed i.e. "Add Cluster" or "Edit Cluster". We don't have an obvious way to do this currently, so it would be good if we could get this added as part of a better UX.

PR #389, opened 2019-07-29
Add Feature: Add and Edit Cluster/Storage with cx check

### LINKED ISSUE #215: Buggy behavior on add cluster
theres still a bit of buggy behavior around the connection piece. When i add a bogus cluster, it shows up as connection failed after adding. then on refresh.. there is not icon state on the left in the data list b/c i think the controller is updating the status. 

Also, I can add duplicate clusters with the same name. Also, sometimes when i add a new cluster, it just doesnt show up in any list until i refresh.

### LINKED ISSUE #229: Implement CheckConnection for clusters
CheckConnection buttons are currently faked out. We need to discuss how we actually want to implement this feature across the app, because we indicate connection status all over. IMO, I think we want to do some kind of shallow validation that says "yes I can talk to this cluster using this token" when a user presses the CheckConnection button. This could be a simple `GET namespaces` succeeds?

The deeper validation I think we should base our "check marks" off of is checking for a Ready condition on the MigCluster/MigStorage object that the controller has confirmed, because it's able to do much deeper checks. These checks are probably too long for the check connection button.

@jwmatthews Curious about your thoughts on this one.

### LINKED ISSUE #230: Implement CheckConnection for storage
CheckConnection buttons are currently faked out. We need to discuss how we actually want to implement this feature across the app, because we indicate connection status all over. IMO, I think we want to do some kind of shallow validation that says "yes I can talk to this bucket using these keys" when a user presses the CheckConnection button. Unlike the cluster connection, I'm not sure what this request would be against s3. Maybe a simple create tmp file, delete tmp file test?

The deeper validation I think we should base our "check marks" off of is checking for a Ready condition on the MigStorage object that the controller has confirmed, because it's able to do much deeper checks. These checks are probably too long for the check connection button.

@jwmatthews also curious about your thoughts on this one.

eriknelson:
I apologize for the size of this one, but didn't see a very clean way to chop it up, so going to try to summarize the high level changes in this one. The overall purpose is to build an improved cluster and storage experience so if the user accidentally enters bad values or something is wrong on the backend, we can alert them, and they can easily continue to edit their objects until it's ready. This also moves us closer to more realistically representing what's actually happening vs. a "Check Connection" button, which was problematic due to its imperative nature.

High level changes:
* Implements the above described add and update experience with status checking
* Ports add/update to cluster/storage sagas and adds watch w/partial updates
* Introduces an AddEditStatus to the cluster/storage sections of the redux tree so you can reliably know the state and mode of the modal
* Converts the Modal and Form to functional components using hooks
* Adds contexts for Polling, Storage, and Cluster related data/fns
* Stops all background polling while modals are open, and starts it again on close
* Removes a lot of dead code

Follow on PRs:
* Needs to be styled according to mocks
* Timeout should allow users to "Try Again"
* There is a LOT of opportunity to consolidate duplicate code. Frankly the entire experience between cluster and storage is almost exactly the same from an abstract standpoint.
* #257 should be easy to knock out with the `AddEditStatus` available now
* Port remaining operations to sagas entirely and remove the operations for consistency

Known issue:
* If the user updates fields that are on objects referenced by Mig objects, and not the Mig object itself, the controller does NOT reconcile, which means if I update something like a clusterUrl to a bad value, it does not update to show me an error condition during the connection check. Need to raise this with the appropriate folks to come up with a solution. To test the connection check of a bad value, edit the AWS bucket region or name on the repo to a bad value. Since these are fields that are directly on the `MigStorage` object, the controller correctly reconciles and reports a critical condition. This can then be edited to be a correct value and it will update to a ready condition.

Closes #230
Closes #229
Closes #215


src/app/home/HomeComponent.tsx
```diff
@@ -280,7 +251,22 @@ class HomeComponent extends React.Component<IProps, IState> {
           <PageSection>
             <Flex justifyContent="center">
               <Box flex="0 0 100%">
-                <DetailViewComponent />
+                <PollingContext.Provider value={{
```
ibolton336: +1 this is awesome

src/app/storage/duck/sagas.ts
```diff
@@ -0,0 +1,265 @@
...
+  AddEditWatchTimeout,
+  AddEditWatchTimeoutPollInterval,
+  AddEditConditionCritical,
+  createAddEditStatusWithMeta,
+  AddEditConditionReady,
+  AddEditDebounceWait,
+} from '../../common/add_edit_state';
+
+function* addStorageRequest(action)  {
+  // TODO: Need to improve this to fall into the failed create state with rollback
```
ibolton336: There is sometimes a bit of a delay when clicking the add button & seeing the button disabled. I think you could add some ui state to show isLoading as soon as the add/update is dispatched for storage/cluster. This could be a candidate for a later issue/refactor.

src/app/cluster/duck/sagas.ts
```diff
@@ -0,0 +1,260 @@
...
+  AddEditWatchTimeout,
+  AddEditWatchTimeoutPollInterval,
+  AddEditConditionCritical,
+  createAddEditStatusWithMeta,
+  AddEditConditionReady,
+  AddEditDebounceWait,
+} from '../../common/add_edit_state';
+
+function* addClusterRequest(action)  {
+  // TODO: Need to improve this to fall into the failed create state with rollback
```
ibolton336: add ui state for isLoading as soon as this is dispatched.

src/app/cluster/duck/sagas.ts
```diff
@@ -0,0 +1,260 @@
...
+    put(alertErrorTimeout('Cluster failed creation'));
+  }
+}
+
+function* watchAddClusterRequest() {
+  yield takeLatest(Creators.addClusterRequest().type, addClusterRequest);
+}
+
+function* updateClusterRequest(action)  {
+  // TODO: Probably need rollback logic here too if any fail
```
ibolton336: add ui state for isLoading as soon as this is dispatched.

src/app/common/add_edit_state.ts
```diff
@@ -0,0 +1,103 @@
...
+    case AddEditState.Pending: {
+      return `Add ${componentType}`;
+    }
+    default: {
+      return `Update ${componentType}`;
+    }
+  }
+};
+
+export const isAddEditButtonDisabled = (
```
ibolton336: could integrate isLoading state here which is dispatched when initial button is clicked from ui to eleminate visual delay.

src/app/storage/duck/sagas.ts
ibolton336: if I fill in the form with bogus values, the add storage button does not disable or give the user any visual feedback 

image

ibolton336: image

storage add is throwing 400 but appears to be connected after refreshing the page

src/app/cluster/duck/sagas.ts
ibolton336: seeing the add fail with "clusterRef must reference a cluster" but shows as connected when I refresh the UI

Issue #256, opened 2019-06-12
When cluster/repo/plan is created, should open that section and close others


eriknelson:
Spawned from #97

PR #402, opened 2019-07-30
Added auto close feature on a second expand

Danil-Grigorev:
This PR will ensure, that only one item is expanded at a particular moment.
Closes https://github.com/fusor/mig-ui/issues/256

### DIFF
--- a/src/app/home/DetailViewComponent.tsx
+++ b/src/app/home/DetailViewComponent.tsx
@@ -22,6 +22,7 @@ import {
 import { StatusPollingInterval } from '../common/duck/sagas';
 import { createAddEditStatus, AddEditState, AddEditMode } from '../common/add_edit_state';
 
+
 interface IProps {
   allClusters: any[];
   allStorage: any[];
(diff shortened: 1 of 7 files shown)


src/app/home/DetailViewComponent.tsx
```diff
@@ -94,11 +111,6 @@ class DetailViewComponent extends Component<IProps, IState> {
     }
   };
 
-  handleWizardToggle = () => {
```
eriknelson: +1

src/app/home/DetailViewComponent.tsx
```diff
@@ -47,29 +47,38 @@ interface IProps {
...
-    isOpen: false,
-    isWizardOpen: false,
-    modalType: '',
-  };
+  constructor (props) {
+    super(props);
+    this.state = {
+      expanded: [],
+      plansDisabled: true,
+      expand: new Array(DataListItems.planList).fill(false),
```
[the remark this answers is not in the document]
eriknelson: Initializing this to `DataListItems.planList` is pretty confusing considering a lot of other places, `planList` is going to refer to all the currently found plans. I would create a specific const that represents the size of the array, like `const NumberOfExpandableRows = 3`. I would define this next to the enum.

src/app/home/DetailViewComponent.tsx
```diff
@@ -152,24 +163,29 @@ class DetailViewComponent extends Component<IProps, IState> {
           <ClusterContext.Provider value={{ watchClusterAddEditStatus }}>
             <ClusterDataListItem
               dataList={allClusters}
-              id="clusterList"
+              id={DataListItems.clusterList}
```
eriknelson: I don't think we want our html ids being something like "0". This should remain a descriptive unique string.

src/app/home/DetailViewComponent.tsx
```diff
@@ -47,29 +47,38 @@ interface IProps {
...
 
 class DetailViewComponent extends Component<IProps, IState> {
-  state = {
-    expanded: [],
-    plansDisabled: true,
-    isOpen: false,
-    isWizardOpen: false,
-    modalType: '',
-  };
+  constructor (props) {
```
ibolton336: Since we are using es7 class props, we dont need the constructor or this.state. https://borgs.cybrilla.com/tils/es7-class-properties/

PR #406, opened 2019-07-30
Logout and refresh token if we get a 401 from the api

### LINKED ISSUE #370: We need to detect when OAuth token is not working
Ran into an issue of:
1) Deploy an env with a specific name, use app.
2) Destroy env, then reprovision with same names, so same DNS names.
3) Log into UI, it's using the cached oauth token in localstorage from prior run.  Token has not expired, but it is from a prev cluster that was destroyed, so reusing token with the new cluster of same DNS names does not work.

We see "Failed to Fetch" on MigPlans, but see 0 Clusters and 0 Storage.
To fix we need to clear localStorage for oauth token.

Recommend we find a way to look to server to tell us if we can use the token and not look for token expiration and need to refresh based on pure client side look at expiration date.  

Screen Shot 2019-07-25 at 10 12 42 AM

### LINKED ISSUE #380: Token expiry refresh is broken
Caught it this morning. I don't know if this is a regression or a bug in the original implementation. Will try to follow up with a an artificial reproducer. Exception:

```
Uncaught TypeError: Cannot read property 'token_endpoint' of null
    at LoginComponent.componentDidUpdate (webpack-internal:///./src/app/auth/LoginComponent.tsx:33)
    at commitLifeCycles (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:17349)
    at commitAllLifeCycles (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18736)
    at HTMLUnknownElement.callCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:199)
    at invokeGuardedCallback (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:256)
    at commitRoot (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18948)
    at eval (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:20418)
    at Object.unstable_runWithPriority (webpack-internal:///./node_modules/scheduler/cjs/scheduler.development.js:255)
    at completeRoot (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:20417)
webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:17117 The above error occurred in the <LoginComponent> component:
    in LoginComponent (created by Context.Consumer)
    in Connect(LoginComponent) (created by Route)
    in Route (created by AppComponent)
    in Switch (created by AppComponent)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Context.Consumer)
    in ConnectedRouterWithContext (created by Context.Consumer)
    in Connect(ConnectedRouterWithContext) (created by AppComponent)
    in ThemeProvider (created by AppComponent)
    in div (created by Context.Consumer)
    in Styled(div) (created by AppComponent)
    in div (created by Context.Consumer)
    in Styled(div) (created by AppComponent)
    in AppComponent (created by Context.Consumer)
    in Connect(AppComponent)
    in Router (created by ConnectedRouter)
    in ConnectedRouter (created by Context.Consumer)
    in ConnectedRouterWithContext (created by Context.Consumer)
    in Connect(ConnectedRouterWithContext)
    in PersistGate
    in Provider

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

eriknelson:
Closes #370
Closes #380

### DIFF
--- a/src/app/auth/LoginComponent.tsx
+++ b/src/app/auth/LoginComponent.tsx
@@ -23,9 +23,13 @@ class LoginComponent extends React.Component<IProps> {
   };
 
   componentDidUpdate = prevProps => {
-    const oauthMeta = this.props.auth.oauthMeta;
     const migMeta = this.props.migMeta;
     const routerLoc = this.props.router.location;
+    const oauthMeta = this.props.auth.oauthMeta;
+    if(!oauthMeta) {
+      this.props.fetchOauthMeta(migMeta.clusterApi);
+      return;
+    }
 
     const freshOauthMeta = !prevProps.auth.oauthMeta && !!oauthMeta;
     const urlParams = new URLSearchParams(window.location.search);
--- a/src/client/client.ts
+++ b/src/client/client.ts
@@ -63,7 +63,7 @@ export class ClusterClient {
         .then(res => resolve(res))
         .catch(err => {
           if(err.response.status === 401) {
-            this.tokenExpiryHandler(this.oldToken())
+            this.tokenExpiryHandler(this.oldToken());
           } else {
             reject(err);
           }
@@ -77,7 +77,7 @@ export class ClusterClient {
         .then(res => resolve(res))
         .catch(err => {
           if(err.response.status === 401) {
-            this.tokenExpiryHandler(this.oldToken())
+            this.tokenExpiryHandler(this.oldToken());
           } else {
             reject(err);
           }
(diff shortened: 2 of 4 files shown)


eriknelson:
This needs some refinement, we need to make the cluster client aware if it's using the oauth token, or using the SA token. This should only happen for the oauth token, which is delivered to the client by the clientfactory only when `hostCluster` is called.

PR #421, opened 2019-08-01
Fix rerender issue

### LINKED ISSUE #378: Dropdowns/popovers/selections flash and disappear inside of tables
Need to optimize rerenders that are created by the background polling. This is causing some user experience headaches

ibolton336:
Closes #378

### DIFF
--- a/src/app/home/HomeComponent.tsx
+++ b/src/app/home/HomeComponent.tsx
@@ -279,7 +279,7 @@ export default connect(
     planStatusCounts: planSelectors.getCounts(state),
     allClusters: clusterSelectors.getAllClusters(state),
     allStorage: storageSelectors.getAllStorage(state),
-    allPlans: planSelectors.getAllPlans(state),
+    allPlans: planSelectors.getPlansWithStatus(state),
     loggingIn: state.auth.loggingIn,
     user: state.auth.user,
     isFetchingClusters: state.cluster.isFetching,
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -75,18 +75,6 @@ const getPlansWithStatus = createSelector(
         latestType,
       };
 
-      //remove controller update fields
-      const { metadata } = plan.MigPlan;
-      if (metadata.annotations || metadata.generation || metadata.resourceVersion) {
-        delete metadata.annotations;
-        delete metadata.generation;
-        delete metadata.resourceVersion;
-      }
-      if (plan.MigPlan.status) {
-        for (let i = 0; plan.MigPlan.status.conditions.length > i; i++) {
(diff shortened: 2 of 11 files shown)


src/app/home/HomeComponent.tsx
```diff
@@ -278,7 +278,7 @@ export default connect(
     planStatusCounts: planSelectors.getCounts(state),
     allClusters: clusterSelectors.getAllClusters(state),
     allStorage: storageSelectors.getAllStorage(state),
-    allPlans: planSelectors.getAllPlans(state),
+    allPlans: planSelectors.getPlansWithStatus(state),
```
eriknelson: This seems dangerous. I don't think we would want to set a property called `allPlans` to a filtered list of plans without a status. Also it's still valid to show plans without status in the data list (dunno if this would actually change that or not).

src/app/plan/duck/selectors.ts
```diff
@@ -78,6 +71,19 @@ const getPlansWithStatus = createSelector(
         hasFailedCondition: hasMigrationError,
         latestType,
       };
+
+      //remove controller update fields
+      const { metadata } = plan.MigPlan;
+      if (metadata.annotations || metadata.generation || metadata.resourceVersion) {
```
eriknelson: Just want to confirm this is basically pulling them out of the state tree in their original form, and then stripping them of these values before passing them onto the components? I don't think we want to change the k8s objects that are our state tree in any way so they purely represent the objects we got out of k8s in their original state.

src/app/home/HomeComponent.tsx
ibolton336: @eriknelson This is actually adding a status field in the selector that the ui is using to calculate status based on other values from each plan object.

src/app/plan/duck/selectors.ts
ibolton336: This is not mutating the original data. IF we want to be able to access these fields in the future, we can add a selector that grabs getOriginalPlans that will allow us to grab the exact object we have in the store.

src/app/plan/duck/selectors.ts
```diff
@@ -134,8 +131,41 @@ const getCounts = createSelector(
   }
 );
 
+const getPlanDiffSelector = createSelector(
```
ibolton336: Optimizing rerenders using reselect memoization -> 
https://github.com/reduxjs/reselect#createselectorinputselectors--inputselectors-resultfunc

-create a selector function that compared the current and old plan objects
- create another selector function that accepts the diff selector as an argument + export this function for use in the UI. 
- if the selectors arguments do not change(in this case the returned value from diff selector function is the argument) the memoized data is passed back from the exported selector to mapstatetoprops. Which won’t trigger a rerender

PR #432, opened 2019-08-06
Finish converting actions to consistent exports

ibolton336:
@eriknelson this finishes up the work that was done recently to change the format of out action export/imports. I noticed the work was not complete for alerts & polling actions.


src/app/common/duck/actions/polling.ts
```diff
@@ -1,35 +1,57 @@
-export const startDataListPolling = (params?: any) => ({
-  type: 'DATA_LIST_POLL_START',
+export const PollingActionTypes = {
```
eriknelson: +1

PR #434, opened 2019-08-06
Improve migration status reporting

djwhatle:
Shows progress for each MigMigration belonging to a MigPlan using the step name and number collected from status conditions.

### DIFF
--- a/src/app/home/components/DataList/Plans/MigrationsTable.tsx
+++ b/src/app/home/components/DataList/Plans/MigrationsTable.tsx
@@ -29,12 +29,15 @@ export default class MigrationsTable extends React.Component<any, any> {
         left: 40px;
       `;
       const type = migration.spec.stage ? 'Stage' : 'Migration';
+      const progressVariant = status.isRunning ? ProgressVariant.info : 
+        (status.isFailed ? ProgressVariant.danger : ProgressVariant.success);
+
       return [
         {
           title: (
             <Flex>
               <StyledBox>
-                <StatusIcon isReady={true} />
+                <StatusIcon isReady={!status.isFailed} />
               </StyledBox>
               <Box>{type}</Box>
             </Flex>
@@ -52,7 +55,7 @@ export default class MigrationsTable extends React.Component<any, any> {
                   value={status.progress} 
                   title={status.stepName} 
                   size={ProgressSize.sm}
-                  variant={status.isRunning ? ProgressVariant.info: ProgressVariant.success}
+                  variant={progressVariant}
                 />
               )}
             </div>
@@ -72,6 +75,7 @@ export default class MigrationsTable extends React.Component<any, any> {
       copied: 0,
       stepName: 'Not started',
       isRunning: false,
+      isFailed: false,
     };
 
     if (migration.status) {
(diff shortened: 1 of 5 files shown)


src/app/plan/duck/selectors.ts
```diff
@@ -64,6 +66,13 @@ const getPlansWithStatus = createSelector(
               return m.status.conditions.some(c => c.type === 'Running');
             }
           }).length;
+
+          if (plan.Migrations[0].status.conditions.length) {
+            latestMigrationStep = plan.Migrations[0].status.conditions.filter(c => {
+              return c.type === 'Running';
+            })[0];
+            console.log(latestMigrationStep = plan.Migrations[0].status.conditions.filter(c => { return c.type === 'Running' })[0])
```
djwhatle: Thinking there's going to be a better way of expressing this cleanly without referencing the 0th index that I don't know exists.

djwhatle:
Todo:
 - Rework [MigrationsTable.tsx ](https://github.com/fusor/mig-ui/blob/21d8f7b9dd269837a8f4118624acce6f25edd5f7/src/app/home/components/DataList/Plans/MigrationsTable.tsx#L62-L125)
 - Figure out how much text is safe to show in the "Last Status" column

src/app/plan/duck/selectors.ts
ibolton336: Using the us find method for this might work

djwhatle:
image

djwhatle:
Current step name and percent complete (current step / total steps) are now shown per Migration in the Migration table.

src/app/home/components/DataList/Plans/MigrationsTable.tsx
```diff
@@ -80,42 +85,45 @@ export default class MigrationsTable extends React.Component<any, any> {
...
-            status.progress = null;
-            break;
-          case 'BackupFailed':
-            status.phase = 'Backup Failed';
-            status.progress = null;
-            break;
-          default:
-            status.phase = migPhase;
-            break;
+        if (migration.status.conditions.length) {
```
ibolton336: We have accumulated some tech debt here. Going forward I would like to propose that we move this type of logic out of the presentation layer & into selectors. This has a few benefits: It is much easier to test (if we ever write tests) and we can leverage reselect to memoize results which allows for fewer re-renders. 

For example, we could create a selector function that takes the current migration state as a param, calculate the migrations status within the function body, & return the derived status for use within the UI.

To do this we would need to pinpoint where the ui state is being updated from the api polling request. From there, instead of passing in the state directly from redux to the UI component, we would pass the redux state-> selector function -> ui component. 

here is an example of what has been done at the plan status level
https://github.com/fusor/mig-ui/blob/787f935b381828a135b340d66d0bdfed006536cf/src/app/plan/duck/selectors.ts#L10

djwhatle:
UI reflects failed migrations

image

djwhatle:
Adding 'wip' back to this, found an issue with controller behavior not matching with what UI allows, working on resolution. https://github.com/fusor/mig-controller/issues/258

djwhatle:
Also need to resolve issue with "Last Status" field displaying incorrect contents in some cases

djwhatle:
End time is now shown correctly for failed migrations, and "Last Status" shows the status accurately when errors occur during migrations.

image

PR #436, opened 2019-08-07
Close plan functionality

### LINKED ISSUE #401: Plans need to be closed on successful final migration and deletion
In order for the controller to do its resource cleanup (registry pods, etc.) and to avoid conflicting plans referencing the same namespaces, the plans need to be closed during deletion, and upon a final migration. The field for closing a plan on the `MigPlan` is `spec.closed`.

* For deletion, we need to mark the field `closed === true`, and then watch the plan for a `type === 'Closed'` condition, which represents the controller acknowledging the closure has completed. Only then can the plan be deleted.
* For final migration, we must be sure not to close the plan prior to final migration being successful. Once that is complete, we should mark the plan closed. Note, in this case it's not necessary to watch for a closed condition, because we are not performing another operation that's dependent on the closure completing. However, we need to also guard against a race condition: if the final migration completes and the plan is marked close, the user cannot delete that plan until the closure has been marked complete.

ibolton336:
Closes #401


src/app/plan/duck/sagas.ts
```diff
@@ -98,29 +99,91 @@ function* watchPlanUpdate() {
...
-function* planDeleteSaga(action) {
+
+function* checkClosedStatus(action) {
+  let planClosed = false;
+  const tries = 0;
+  const TicksUntilTimeout = 20;
+  while (!planClosed) {
+    if (tries < TicksUntilTimeout) {
+      const getPlanResponse = yield call(getPlanSaga, action.planName);
+      const MigPlan = yield getPlanResponse.data;
```
eriknelson: Don't think this needs to be yielded

PR #441, opened 2019-08-13
Pv details json view & table cleanup

### LINKED ISSUE #312: Implement 'View Details' link in PV Discovery

ibolton336:
Closes #312


src/app/plan/components/Wizard/VolumesTable.tsx
```diff
@@ -30,12 +40,19 @@ const VolumesTable = (props): any => {
...
+  const getPVResources = (pvList = [], clusterName = '') => {
+    getPVResourcesRequest(pvList, clusterName);
   };
 
   useEffect(() => {
     if (currentPlan) {
       const discoveredPersistentVolumes = currentPlan.MigPlan.spec.persistentVolumes || [];
+      //get resources for pv details
+      getPVResources(discoveredPersistentVolumes, values.sourceCluster);
+      //
```
eriknelson: Could remove this comment

src/app/plan/duck/sagas.ts
```diff
@@ -115,12 +104,49 @@ function* planDeleteSaga(action) {
   }
 }
 
+function* getPVResourcesRequest(action) {
+  const state = yield select();
+  const client: IClusterClient = ClientFactory.forCluster(action.clusterName, state);
+  try {
+    const resource = new CoreClusterResource(CoreClusterResourceKind.PV);
+    const pvResourceRefs = yield all(action.pvList.map(pv => {
```
eriknelson: This is going to fail all of them if any of them fail. I think we want Promise.allSettled again here. I would pull in the Q npm module for this since they have the utilities we're after, and I need to use it for rollback as well.

Off the top of my head, I think reasonable behavior for failures would be to just dispatch the successful ones, and then in the component, if the tooltip is opened for a failed one, and we don't have the PV contents, we can render in an error into the tooltip like `PV Not Found`.

src/client/resources/common.ts
```diff
@@ -42,7 +42,7 @@ export abstract class ClusterResource {
   public listPath(): string {
     return ['/apis', this.gvk().group, this.gvk().version, this.gvk().kindPlural].join('/');
   }
-  public namedPath(): string {
+  public namedPath(name?): string {
```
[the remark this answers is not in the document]
ibolton336: haha yeah I was wondering about that also. I think since we never requested a named  core resource we never ran into it? idk

src/app/plan/duck/sagas.ts
ibolton336: Yeah I started going down the rabbit hole of when to make this request last night. Thought about making the request on the actual click of the view details button. But that would present a different set of issues. I will go with the error message for now since it doesnt seem to be a very resource intensive request.

PR #442, opened 2019-08-13
Support for migration on mocked data

Danil-Grigorev:
This change for mocked client adds ability to create and delete resources. `Local Storage` values are getting updated too.


src/client/kube_store/index.ts
```diff
@@ -43,166 +39,72 @@ export default class KubeStore {
...
-        }
-      } else {
-        // Extension ClusterResource
-        key = `apis/${resource.gvk().group}/${resource.gvk().version}/${resource.gvk().kindPlural}`;
-        if (this.data()[key]) {
-          mockedObject = {...this.data()[key][name]};
-        }
-      }
-    }
+    const apiKey = resource.listPath().substr(1);
```
jwmatthews: +1 

I like how you cleaned this up

jwmatthews:
This is looking really good!

I just tested it out and had success doing a Stage/Migrate of the plan that is already in the mocks.

I then went through and created a new plan, was able to see mocked namespaces and pv selections, completed plan.  Then was able to do a Stage.

Seeing progress tick as updates in status come in.  I did see an error of 'Failed to update Plan', that showed up so probably need to test some more on that.

Overall the behavior of this is spot on for what we are looking for.

config/webpack.dev.js
```diff
@@ -69,6 +69,22 @@ if (devMode === 'local') {
   plugins.push(
     new webpack.NormalModuleReplacementPlugin(/client_factory.ts/, 'client_factory.mock.ts')
   );
+  plugins.push(
+    new webpack.NormalModuleReplacementPlugin(/plan\/duck\/operations\.ts/, res => {
```
jwmatthews: This is something we'd like to avoid.

Would like to keep the mock interface at the client level.

config/webpack.dev.js
```diff
@@ -69,6 +69,22 @@ if (devMode === 'local') {
...
+      if (res.rawRequest === '../operations') return;
+      res.request = res.request.replace(/operation.\.ts/, 'mock/operations.mock.ts');
+      res.resource = res.resource.replace(/operation.\.ts/, 'mock/operations.mock.ts');
+      res.resourceResolveData.path = res.resourceResolveData.path.replace(/operation.\.ts/, 'mock/operations.mock.ts');
+    })
+  );
+  plugins.push(
+    new webpack.NormalModuleReplacementPlugin(/(?:cluster|storage|plan)\/duck\/sagas\.ts/, res => {
```
jwmatthews: This is something we'd like to avoid.

Would like to keep the mock interface at the client level.

Danil-Grigorev:
Agree on point with merging to a new branch. The logic could be later migrated to the `kube_store` with almost/no changes. Right now, only points of interception with `sagas` and `operations` code, are those captured actions, which should dispatch an original action and mocked ones - https://github.com/fusor/mig-ui/pull/442/files#diff-60aa9b20bccfafd269e83e6c3a6c7cb8R30-R31

src/client/kube_store/index.ts
```diff
@@ -43,166 +39,226 @@ export default class KubeStore {
...
+      obj['status']['conditions'] = [{
+        lastTransitionTime: new Date().toUTCString(),
+        type: statusType,
+      }];
+      this.updateMockedData(apiKey, resourceName, obj);
+    }, timeout);
   }
 
-  private updateMockedData(key, name, updatedObject) {
+  private discoverPVsMock = (apiKey, resourceName, migPlan) => {
```
eriknelson: :+1:

src/client/kube_store/index.ts
```diff
@@ -43,166 +39,226 @@ export default class KubeStore {
...
-    newData['clusters'][this.clusterName][key][name] = updatedObject;
-    localStorage.setItem(localStorageMockedDataKey, JSON.stringify(newData)); 
+    newData.clusters[this.clusterName][apiKey][resourceName] = updatedObject;
+    localStorage.setItem(LocalStorageMockedDataKey, JSON.stringify(newData));
+    if (['MigCluster', 'MigStorage'].includes(updatedObject.kind)) {
+      this.setResourceStatus(apiKey, resourceName, updatedObject, 'Ready');
+    }
+  }
+
+  private createMockedData(apiKey, resourceName, obj) {
```
eriknelson: This is great, I really like how you broke out the steps into functions.

eriknelson:
Confirmed this is working great with a simple `yarn start`. The only thing I noticed was that I couldn't add a new cluster and try to migrate from that as a source, however, I think that's something we can reasonably add if we want to in another PR. The UI also handles this gracefully with a simple error alert.

PR #444, opened 2019-08-14
Added ability to gather mocked data from remote clusters

Danil-Grigorev:
Resolves secret references from migclusters, and collects `namespaces` and `storage classes` for those clusters.


jwmatthews:
I'll test this out on Friday, quick look all looks good, I'd like some time to try this out in my environment and then will ack/merge it in.

scripts/gather-mocked-data.js
```diff
@@ -98,24 +104,47 @@ function writeData(data) {
...
+      .catch(err => {
+        console.error(err);
         process.exit(1);
       });
   });
+
   //  Reducing to a single promise that has collected all of the results.
   //  Transform from array to a dict that matches the layout of how we want to store the data
-  return all.reduce((promiseChain, currentTask) => {
+  return namespaced.reduce((promiseChain, currentTask) => {
```
eriknelson: This ugly nesting should be cleaned up with async/await, but I think it's fine to merge this as is and file an issue to improve stuff.

PR #448, opened 2019-08-14
Adds support for non-AWS S3 providers

### LINKED ISSUE #422: Support usage of NooBaa and Minio as well as AWS S3
Need to verify we are capturing correct fields to work with generic S3 providers.

eriknelson:
Closes #422

### DIFF
--- a/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
+++ b/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
@@ -8,6 +8,7 @@ import KeyDisplayIcon from '../../../common/components/KeyDisplayIcon';
 import HideWrapper from '../../../common/components/HideWrapper';
 import utils from '../../../common/duck/utils';
 import storageUtils from '../../duck/utils';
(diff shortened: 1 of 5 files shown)


src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
```diff
@@ -96,6 +98,20 @@ const InnerAddEditStorageForm = ({ values, touched, errors, ...props }) => {
           <FormErrorDiv id="feedback-bucket-name">{errors.bucketName}</FormErrorDiv>
         )}
       </FormGroup>
+      <FormGroup label="S3 URL" fieldId={s3UrlKey}>
```
dymurray: Actually, thoughts on changing this to `S3 Endpoint`? Seems to be the common language I see across services using the S3 API.

eriknelson: No problem

eriknelson: Updated

PR #466, opened 2019-08-21
Implement cluster addition protections

eriknelson:
* Implements rollback - if some cluster objects have failed, must
rollback those that actually failed
* Implements protection against trying to create objects that already
exist

### DIFF
--- a/src/app/cluster/duck/sagas.ts
+++ b/src/app/cluster/duck/sagas.ts
@@ -30,9 +30,9 @@ import {
   AddEditConditionReady,
   AddEditDebounceWait,
 } from '../../common/add_edit_state';
+import Q from 'q';
 
 function* addClusterRequest(action) {
-  // TODO: Need to improve this to fall into the failed create state with rollback
   const state = yield select();
   const { migMeta } = state;
   const { clusterValues } = action;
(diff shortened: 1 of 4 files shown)


src/app/cluster/duck/sagas.ts
```diff
@@ -65,15 +65,91 @@ function* addClusterRequest(action) {
...
+      };
+
+      // The objects that need to be rolled back are those that were fulfilled
+      const rollbackObjs = clusterAddResults.reduce((rollbackAccum, res) => {
+        return res.state === 'fulfilled' ?
+          [...rollbackAccum, { kind: res.value.data.kind, name: res.value.data.metadata.name }] :
+          rollbackAccum;
+      }, []);
+
+      const rollbackResults = yield Q.allSettled(rollbackObjs.map(r => {
```
ibolton336: 👍

ibolton336: Only 2 comments with this:

- Check connection button shows ready after clicking 2 times. Any changes we can make this a better experience?

- Are we changing the check connection button to check with the new values entered in the form? Or is the current behavior ok?

PR #469, opened 2019-08-22
Removed usage of put request in favor of patch

### LINKED ISSUE #468: Plan watcher is using put request instead of patch
The current implementation, which updates the migPlan state during different stages of the migration is based on `PUT` requests, which rewrites existing plan completely. This behaviour has some issues, one of them is that the status field get's purged every time when such update is executed.

The idea is to do the same set of actions, but by using the `PATCH` request instead, which will do it more mildly, and avoid described issues.

This also will allow to use existing resource as a source of truth, and will allow to gain more data about the plan for the mocked mig-controller in #442. In some occasions to delete the migplan, and decide if it needs a mocked mig-controller response, there is a need to look into the `status` of the migplan. When this information gets lost, mocks can't simulate constant watching and updates on the definition as the controllers do, which causes issues.

Danil-Grigorev:
Closes #468 by eliminating usage of put requests while updating existing MigPlan.

### DIFF
--- a/src/app/plan/duck/sagas.ts
+++ b/src/app/plan/duck/sagas.ts
@@ -61,16 +61,16 @@ function* getPlanSaga(planName) {
     throw err;
   }
 }
-function* putPlanSaga(getPlanRes, planValues) {
+function* patchPlanSaga(getPlanRes, planValues) {
   const state = yield select();
   const migMeta = state.migMeta;
   const client: IClusterClient = ClientFactory.hostCluster(state);
   try {
-    const updatedMigPlan = updateMigPlanFromValues(getPlanRes.data, planValues);
-    const putPlanResponse = yield client.put(
+    const patch = updateMigPlanFromValues(getPlanRes.data, planValues);
+    const putPlanResponse = yield client.patch(
       new MigResource(MigResourceKind.MigPlan, migMeta.namespace),
       getPlanRes.data.metadata.name,
-      updatedMigPlan
+      patch
     );
     yield put({ type: PlanActionTypes.UPDATE_PLAN, updatedPlan: putPlanResponse.data });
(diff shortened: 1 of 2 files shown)


eriknelson:
Have you actually tested this? I don't think this is as simple as just swapping the method call. We need to make sure the patch object is correct.

eriknelson:
I also don't think we actually have a consensus if it's actually **wrong** to use PUT here. The convenience of UI mocks shouldn't be the driving factor. The question is really which one is the more correct choice from a kube API perspective?

Danil-Grigorev:
I tested this on several migrations, where I tried the functionality on a plan closure and PV discovery, not noticing any issues. Basically previous implementation was doing the patching already, but did it locally, by copying all values from the existing resource on a new resource. https://github.com/fusor/mig-ui/blob/511f0d55838d4772cd31b3f897746f5b5be51ecd/src/client/resources/conversions.ts#L244-L247

The kuberenetes documentation specifies that the existing deployment objects are updated by the patch request. https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/ The migPlan is close at it's functionality to a Deployment, it watches the state of migMigrations and do some functionality specified by it's definition. There is no info about usage of patch for other resources, but I don't see if such action is prohibited by some rule or kubernetes anti-pattern.

eriknelson:
Labelling this WIP for a couple reasons:
* I think the PUT vs. PATCH conversation needs some more discussion with the controller team.
* This change as it stands isn't really correct for patch. The patch payload should contain only the fields that need to be updated, not the whole object as you would with a PUT.

eriknelson:
Discussed with @Danil-Grigorev and we're closing this due to reasons discussed on #468

Issue #465, opened 2019-08-21
PVs Moved and PVs Copied fields not updated under Migration Status


jwmatthews:
Click on 'Migrations' after doing a stage/migrate.
We should see the number of PVs copied/moved displayed, but currently shows 0

<img width="1680" alt="2019-08-21 09 03 11" src="https://user-images.githubusercontent.com/206875/63434166-8a38f000-c3f2-11e9-8f8c-68c5454aa958.png">

PR #472, opened 2019-08-23
Added list of copy/move(ed) PVs for migration details

Danil-Grigorev:
- PVs moved/copied is connected to the plan specification.
- Some refactoring in the component.
image

Closes https://github.com/fusor/mig-ui/issues/465


src/app/home/components/DataList/Plans/MigrationsTable.tsx
```diff
@@ -77,7 +44,10 @@ export default class MigrationsTable extends React.Component<any, any> {
       isFailed: false,
       isSucceeded: false,
     };
-
+    if (plan.spec.persistentVolumes && !migration.spec.stage) {
```
eriknelson: PVs can be still be copied during a stage, just not moved. This condition doesn't cover that case.

PR #506, opened 2019-08-30
convert to close to patch

### LINKED ISSUE #496: Convert plan close on final migration from PUT to PATCH request
Saw a 409 likely due to bad PUT logic once a final migration finished. This update should probably be PATCH based anyway.

ibolton336:
Closes #496

### DIFF
--- a/src/app/plan/duck/sagas.ts
+++ b/src/app/plan/duck/sagas.ts
@@ -66,6 +66,26 @@ function* getPlanSaga(planName) {
     throw err;
   }
 }
+function* patchPlanSaga(planValues) {
+  const state = yield select();
+  const migMeta = state.migMeta;
+  const client: IClusterClient = ClientFactory.hostCluster(state);
+  try {
+    const getPlanRes = yield call(getPlanSaga, planValues.planName);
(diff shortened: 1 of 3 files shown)


src/app/plan/duck/sagas.ts
```diff
@@ -66,6 +66,26 @@ function* getPlanSaga(planName) {
     throw err;
   }
 }
+function* patchPlanSaga(planValues) {
+  const state = yield select();
+  const migMeta = state.migMeta;
+  const client: IClusterClient = ClientFactory.hostCluster(state);
+  try {
+    const getPlanRes = yield call(getPlanSaga, planValues.planName);
+    const updatedMigPlan = updateMigPlanFromValues(getPlanRes.data, planValues);
```
eriknelson: Danil did something similar with his PR -- I think the only thing that should be in the payload should be the patched value, not the rest of the object that the conversion returns. It was written for PUT, so it returns everything. We only want to include a JSON merge patch body that updates the closed value to true.

https://tools.ietf.org/html/rfc7386

ibolton336: @eriknelson just updated

PR #510, opened 2019-08-30
chore(mig-ui): add sass dependencies, loaders and partials

mattnolting:
- add sass dependencies and loaders
- add css for widget card layout, markup to follow
- update react and react-table

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -95,8 +95,15 @@ const webpackConfig = {
         use: 'ts-loader',
       },
       {
-        test: /\.css$/,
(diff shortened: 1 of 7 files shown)


ibolton336:
@mattnolting thanks for doing this - we will want to merge this after we get a stable version for our upcoming beta. Need to verify that the version upticks aren't going to cause any issues with our PF4 implementation

mattnolting:
@ibolton336 how about I close this in favor of a PR that adds only the sass dependency, which would leave `pf-react` and `pf-table` version unaltered?

PR #519, opened 2019-09-03
Add sass dependency and handlers

mattnolting:
- Add sass dependency only


ibolton336:
need to investigate what this is changing with types. Seeing a warning thrown with the migrations history table

image

PR #526, opened 2019-09-04
only update current plan if plan changes. Fixes rerender regression

### LINKED ISSUE #378: Dropdowns/popovers/selections flash and disappear inside of tables

ibolton336:
Closes #378


eriknelson:
Tested a couple migrations with this and looks good

PR #527, opened 2019-09-04
feature(icons): add icon support

mattnolting:
adds icon css and html support


eriknelson:
Tested a bunch of different angles and didn't see anything out of the ordinary. Thank you @mattnolting!

PR #533, opened 2019-09-05
refactor(home): update home layout

mattnolting:


eriknelson:
Merge conflict on this one

ibolton336:
image

We received some initial help on the card component format from UXD. @mcoker and @kyle gave us mockups which we implemented for these cards. 

I know the most recent feedback requested a change of the hierarchy  ( https://marvelapp.com/66cj5ga/screen/60982350 ) , but I think we might have to give this more thought. Feels like a step back IMO

mattnolting:
@ibolton336 Looks like the className changed and the cards lost styling. Updated that and added support for `.pf-c-page__header` background-color. 

@eriknelson conflicts resolved.

PR #536, opened 2019-09-05
feat(datalist-badges): add badges to datalist names

mattnolting:
adds generic badge, mig-ui team will needs to update badge count


eriknelson:
Marking as `do not merge`; we'd like to bring this in but definitely need to get this hooked up.

PR #538, opened 2019-09-05
Discontinue using registry cluster

### LINKED ISSUE #296: Remove usage of cluster-registry
We plan to discontinue usage of cluster-registry
See below issue from mig-controller: 
https://github.com/fusor/mig-controller/issues/173

eriknelson:
**Depends on backend support, PR posted here: https://github.com/fusor/mig-controller/pull/267**
**Do not merge until both are coordinated!**

Closes #296

This PR removes our dependency on the ClusterRegistry object, and moves a cluster's URL to the `MigCluster` object under the field `url`.


eriknelson:
I've tested this adding, editing, staging, and migrating. Needed to add permissions to the controller but that was unrelated. I have an image posted to `quay.io/eriknelson/mig-ui:nocr` with a build, would be great if some folks could try this out.

PR #539, opened 2019-09-05
refactor(wizard-layout): update step layouts

mattnolting:
Updates step layouts (not final results). Still work to be done in some details.


eriknelson:
@ibolton336 you've spent a lot of time in this area, can you take a look?

PR #602, opened 2019-10-26
Add check for namespace limit

ibolton336:


src/app/plan/duck/utils.ts
```diff
@@ -1,14 +1,22 @@
 import moment from 'moment';
 
-const getPlanPVs = plan => {
-  const statusObj = { success: null, error: null };
+const getPlanPVsAndCheckConditions = plan => {
+  const statusObj = { success: null, error: null, errorText: null };
   const PvsDiscoveredType = 'PvsDiscovered';
-
+  const NamespaceLimitErrorType = "NamespaceLimitExceeded";
```
dymurray: Travis wants this to be single quoted

PR #611, opened 2019-11-06
migcluster: add 'insecure' and 'caBundle' fields to add/edit form

### LINKED ISSUE #316: Support passing in CA cert for remote clusters to support mig-controller to use https to server side communication to MigClusters
Related to:
https://github.com/fusor/mig-controller/issues/91

mansam:
Adds an option for disabling SSL verification and an option to provide a base64 encoded caBundle to the migcluster add/edit form. This provides the necessary front end support for https://github.com/fusor/mig-controller/pull/347.

migcluster-ca-bundle-insecure
Fixes #316

### DIFF
--- a/src/client/resources/conversions.ts
+++ b/src/client/resources/conversions.ts
@@ -36,9 +36,11 @@ export function createMigCluster(
   name: string,
   namespace: string,
   clusterUrl: string,
-  tokenSecret: any,
   isAzure: boolean,
-  azureResourceGroup: string
+  azureResourceGroup: string,
+  insecure: boolean,
+  caBundle: string,
+  tokenSecret: any,
 ) {
   let specObject;
   if (isAzure) {
@@ -46,6 +48,8 @@ export function createMigCluster(
       azureResourceGroup,
       isHostCluster: false,
       url: clusterUrl,
+      insecure,
+      caBundle: caBundle || null,
       serviceAccountSecretRef: {
         name: tokenSecret.metadata.name,
         namespace: tokenSecret.metadata.namespace,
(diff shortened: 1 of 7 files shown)


src/client/resources/conversions.ts
```diff
@@ -55,12 +59,16 @@ export function createMigCluster(
...
+export function updateMigCluster(
   clusterUrl: string,
+  insecure: boolean,
+  caBundle: string,
 ) {
   return {
     spec: {
       url: clusterUrl,
+      insecure,
+      caBundle: caBundle || null,
```
mansam: The backend returns a 422 if this value is the empty string, since it's apparently expecting either null or a base64 encoded string that it can fit into a `[]byte`. Fixing it this way seems sort of ugly, not sure if there's a better approach.

mansam:
@eriknelson I've got a text field here at the moment as the input for the CA bundle, but I'd think either a text area to allow pasting the PEM cert without having to b64 encode it first or a file upload would be better suited. Anyone have any opinions on that?

vconzola:
@mansam There's a PatternFly component that allows either text copy or file upload but it hasn't been implemented in PF4 yet. I'll talk to @ibolton336 when I get back in the U.S. about redesigning in PF4.

vconzola:
@ibolton336 Here's a [link](https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/common/forms/TextFileInput/TextFileInput.js) to the code for the above mentioned component. Do you think you can rebuild this with PF4 styles?

ibolton336:
> @ibolton336 Here's a [link](https://github.com/ManageIQ/manageiq-v2v/blob/master/app/javascript/react/screens/App/common/forms/TextFileInput/TextFileInput.js) to the code for the above mentioned component. Do you think you can rebuild this with PF4 styles?

https://codesandbox.io/s/lkkjpr5r7 I think if we can just use a simple HTML5 file input element, we should be good (as shown in this link within a formik form). Unless we need drag and drop functionality, this should be enough IMO. & we can style the button & text to be pf4 compliant.

PR #610, opened 2019-11-06
Fixed unselection of namespaces after Select All click

Danil-Grigorev:
Deselecting any of the namespaces after `select all` was clicked didn't change selected namespaces in the migPlan spec.

PR #590, opened 2019-09-19
Edit/undo wizard functionality

ibolton336:
Closes #311 
--Improves naming convention for open/toggle booleans
-- Refactor PV discovery poll to use plan status poll
  - In WIzardComponent, status polling & update request is triggered when entering the results step
  - In VolumesForm, status polling & update request is triggered when loading the volumes component

DONE: conversions.ts needs to be updated to differentiate when namespace list has changed. Need to figure out how to know when pv discovery has re-run & been completed successfully

PR #612, opened 2019-11-07
Plan undo, reconcile driven plan updates.

### LINKED ISSUE #614: Plan status not being updated properly on results page
When arriving at the results page, the plan status is cached & does not indicate if there are any changes from the cached status. Need to set the currentPlan status to pending when arriving at this page to fix this bug

Danil-Grigorev:
Plan is to leverage #590 
Fixes "Bug 1758336 - [MIG UI] Unable to edit namespace selection." in combination with #610 
Closes #614
Changes:
- Undo functionality, reconcile watch.
- Resource version polling. Idea is to set it as a preceding step before viewing/operating on any plan's state values.
- Put instead of patch on migPlan creation.
- Improved performance
- Added resourceVersion driven checks.
- Added deduplication on plan update.
- Refactored logic of migration plan creation/editation + closure.

### DIFF
--- a/src/app/plan/duck/actions.ts
+++ b/src/app/plan/duck/actions.ts
@@ -34,6 +34,8 @@ export const PlanActionTypes = {
   PLAN_RESULTS_REQUEST: 'PLAN_RESULTS_REQUEST',
   INIT_STAGE: 'INIT_STAGE',
   INIT_MIGRATION: 'INIT_MIGRATION',
+  PLAN_RECONCILE_POLL: 'PLAN_RECONCILE_POLL',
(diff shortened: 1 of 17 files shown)


src/app/plan/duck/actions.ts
```diff
@@ -192,6 +191,15 @@ const initMigration = (planName: string) => ({
   planName,
 });
 
+const planReconcilePolling = (planValues) => ({
```
ibolton336: For now,  @eriknelson and I committed to using the generations field to determine if reconcile is complete. There is some great stuff in this PR but i think we should open a smaller PR to incrementally introduce some of these enhancements.

PR #616, opened 2019-11-19
Only update redux tree if the clusters have actually been updated

ibolton336:
Improves performance by limiting rerenders from background polling


ibolton336:
Worked around this using polling context to start and stop polling more granularly.

PR #636, opened 2019-12-10
Initial implementation for discovery client

Danil-Grigorev:
```js
const discoveryClient: IDiscoveryClient = ClientFactory.discovery(state);
const namespaces = new NamespaceDiscovery(state.migMeta.namespace, clusterName);
const res = await discoveryClient.get(namespaces);
```

This PR introduces client for Discovery resources. The `config.dev.json` have a new change, and will need another field for `discoveryApi`, which will be used for namespace discovery:
```json
  "discoveryApi": "http://discovery-api.supercluster.example.com",
```

### DIFF
--- a/config/config.dev.json.example
+++ b/config/config.dev.json.example
@@ -6,5 +6,6 @@
   "redirectUri": "http://localhost:9000/login/callback",
   "namespace": "openshift-migration",
   "configNamespace": "openshift-config",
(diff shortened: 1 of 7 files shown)


eriknelson:
Closing in favor of https://github.com/fusor/mig-ui/pull/640, which has superseded this.

PR #640, opened 2019-12-11
Discovery api/frontend integration

Danil-Grigorev:
Based on #636


src/client/resources/common.ts
```diff
@@ -50,3 +56,41 @@ export abstract class ClusterResource {
 function namedPath(listPath, name) {
   return [listPath, name].join('/');
 }
+
+export abstract class ClusterDiscoveryResource {
```
eriknelson: RC: We don't want to call this `ClusterDiscoveryResource`. That could be confusing when seen in the context of kubernetes resources, which are usually prefixed as `Cluster` resources when they are cluster-scoped and non-namespaced.

I think simply `DiscoveryResource` as an abstract class works here.

src/client/resources/common.ts
```diff
@@ -50,3 +56,41 @@ export abstract class ClusterResource {
...
+
+  private _type: string;
+  private _cluster: string;
+
+  public abstract discoveryParameters(): IDiscoveryParametrs;
+  public abstract for(): string;
```
eriknelson: RC: The name of this method should be more descriptive about what it's abstracting. It looks like this is expected to return the `resourcePath`, or something similar.

src/client/resources/discovery.ts
```diff
@@ -0,0 +1,26 @@
+import { ClusterDiscoveryResource, IDiscoveryParametrs } from './common';
+
+export class NamespaceDiscovery extends ClusterDiscoveryResource {
+  private _discoveryParameters: IDiscoveryParametrs;
+
+  constructor(cluster: string, { wait = -1, offset = -1, limit = -1 } = {}) {
```
eriknelson: RC: See magic number comment, should be a const defined somewhere and used here to default it.

src/client/resources/common.ts
```diff
@@ -50,3 +56,41 @@ export abstract class ClusterResource {
...
+  }
+
+  public discoveryType() { return this._type + '/'; }
+
+  public discoveryCluster() {
+    return ['clusters', this._cluster].join('/');
+  }
+
+  public parametrizedPath(params = {}) {
+    if (this.discoveryParameters().wait !== -1) {
```
eriknelson: RC: `-1` here is a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) and an anti-pattern. It would be better to define a const somewhere like `UNSPECIFIED_DISCOVERY_PARAM`. Could even go as far as to typedef the int, but I don't think that's necessary.

eriknelson:
Marking as DO NOT MERGE because there seems to still be some issues with https. Going to test and run this now and see if I can get to the bottom of that.

src/client/resources/common.ts
```diff
@@ -11,6 +11,12 @@ export interface IGroupVersionKindPlural {
   kindPlural: string;
 }
 
+export interface IDiscoveryParametrs {
+  wait: number;
```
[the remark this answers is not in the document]
Danil-Grigorev: Fixed.

src/app/common/components/AlertModal.tsx
```diff
@@ -0,0 +1,46 @@
+import styled from '@emotion/styled';
```
Danil-Grigorev: Part of https://github.com/fusor/mig-ui/pull/645

src/app/logs/components/LogsContainer.tsx
```diff
@@ -21,86 +19,60 @@ import {
...
+      const element = document.createElement('a');
+      element.href = archive;
+      element.download = 'logs.zip';
+      setDownloadLink(element);
+    }
+  }, [archive]);
+
+  useEffect(() => {
+    if (downloadLink) { downloadArchive(downloadLink); }
+  }, [downloadLink]);
```
Danil-Grigorev: Making sure operations will happen only on mounted component

src/app/logs/duck/sagas.ts
```diff
@@ -1,196 +1,130 @@
...
+  const state = yield select();
+  const discoveryClient: IDiscoveryClient = ClientFactory.discovery(state);
+  const report: IPlanLogSources = action.report;
+  try {
+    const archive = new JSZip();
+    const logPaths = flatten(Object.values(ClusterKind).map(
+      src => report[src].map(pod => pod.log)));
+
+    const logs = [];
+    for (const log of logPaths) {
```
Danil-Grigorev: Necessary evil to do it sequentially, as concurrent requests broke the discovery a bit @jortel

src/app/plan/duck/sagas.ts
```diff
@@ -345,14 +349,11 @@ function* planCloseAndDeleteSaga(action) {
 
 function* getPVResourcesRequest(action) {
   const state = yield select();
-  const client: IClusterClient = ClientFactory.forCluster(action.clusterName, state);
+  const discoveryClient: IDiscoveryClient = ClientFactory.discovery(state);
```
eriknelson: +1, nice

src/client/client.ts
```diff
@@ -150,29 +150,29 @@ export class ClusterClient {
   }
 
   private checkExpiry() {
-    if(this.isTokenExpired() && this.tokenExpiryHandler) {
```
eriknelson: Q: What's the reason to pull this?

src/client/client_factory.ts
```diff
@@ -45,6 +53,22 @@ export const ClientFactory = {
     const newClient = new ClusterClient(clusterApi, accessToken, false /*isOauth*/, customResponceType);
     return newClient;
   },
+  discovery: (state: any, customResponceType: ResponseType = 'json') => {
```
eriknelson: RC: Should be `customResponseType`, spelling

src/client/resources/convension.ts
```diff
@@ -0,0 +1,22 @@
+export type INamespaceList = string[];
```
eriknelson: I'm not sure about where this lives as I mentioned, let's talk through it.

src/client/client.ts
Danil-Grigorev: Currently we have two clients, discovery and native kubernetes one, both are using same logic for token expiry handler. Actually, just realizing that from now on they both will always use the oauth token, as we are working with single `host` cluster. Will adjust the logic a bit, thanks for noting.

src/client/resources/convension.ts
Danil-Grigorev: Really it is just a list of strings, let's remove this.

src/app/logs/components/LogBody.tsx
```diff
@@ -11,17 +11,27 @@ import {
   Title
 } from '@patternfly/react-core';
 import LogItem from './LogItem';
+import { connect } from 'react-redux';
 import { Spinner } from '@patternfly/react-core/dist/esm/experimental';
-// const LogItem = lazy(() => import('./LogItem'));
+import { IPlanLogSources } from '../../../client/resources/convension';
```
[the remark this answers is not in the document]
eriknelson: Plan is to move these types to `discovery.ts`

PR #653, opened 2020-01-14
Bug 1790966 - [MIG-UI] Edit plan shows wrong information

Danil-Grigorev:


eriknelson:
I can't reproduce the original bug on our latest images without this change: https://bugzilla.redhat.com/show_bug.cgi?id=1790966

I don't want to accept this unless we're sure that it's repairing a reproducible bug, so I'm going to mark this DNM until that's clearly demonstrable.

eriknelson:
This isn't a relevant change to the posted bz; I've investigated and confirmed it's of a much wider scope that has regressed plan editing. Release is being delayed to correct this with proper QE testing.

PR #666, opened 2020-02-12
Refactor: styled components(emotionJS) --> Css Modules using Sass pre-processor

### LINKED ISSUE #287: Remove usage of emotion inline styles in favor of separate SASS files
We've been hitting an issue with PF4 for a while now where they are using an older emotion dep than is currently available. #284 is an example of some real problems where the types simply aren't compatible between our more recent version and theirs; we can't style PF4 components with the latest emotion.

It doesn't sound like PF4 intends to update their dep any time soon, so we need a longer term solution to try to break this coupling. After discussing our options with @ibolton336, we decided to drop `@emotion` styles all together in favor of separate SASS stylesheets that can live alongside their relevant components. The benefit here is we remove any dependency hell risk and can style anything we like with simple sheets.

The work includes:
* Remove emotion dependencies
* Converting all of our custom styles to SASS files and importing them
* Updating the webpack builds to be able to handle this correctly

Question:
* Do we prefer .scss or .sass format (Ex: https://responsivedesign.is/articles/difference-between-sass-and-scss/)

### LINKED ISSUE #608: Logo not displaying in production on logs page

### LINKED ISSUE #670: "Migrate" window text
Screenshot: https://github.com/fusor/mig-agnosticd/blob/master/demos/2019_RHTE/labs/screenshots/lab5/cam-quiesce.png

"Halting" is not normally used in OpenShift docs. "Stopping" is the preferred term.
I suggest changing the text in this window:

Applications [or pods or processes] on the source cluster are stopped before migration begins and restarted after migration. Persistent volumes associated with the migrated projects are moved or copied to the target cluster, according to the migration plan.

Check box: Do not stop applications on the source cluster during migration.

ibolton336:
This PR introduces Sass/Css modules as a replacement for emotionJS & styled components. It also includes: 
- Closes #608 
- Closes #287 
- Closes #670 
- Update @patternfly/core & @patternfly/table versions 
- Update production ejs template to import css from yarn build command inside the container. 
- Convert beta PF4 flex & flex-item classNames to new PF4 react components
- Webpack configuration for CSS/Sass modules in dev mode & production

To style a component:
1) Inside the .jsx component file:
`const styles = require('./AddEditStorageForm.module');`

2) Create a new file in the component directory called: 
`AddEditStorageForm.module.scss`

3) Add style to .scss module file:
`.testStyleForComponent{ color: red}`

4) In .jsx component file, you can style your JSX using the:
`<GridItem className={styles.testStyleForComponent}>`


eriknelson:
image
Seeing a regression with the plan validation not being centered in the wizard.

eriknelson:
image

Spacing on the migrate prompt is a little odd as well. Think this needs some repair.

eriknelson:
image

"Select Pod Source" needs some top margin on the logs screen, not sure if this is a regression or has always been there.

PR #691, opened 2020-02-21
Use helperText props in FormGroup for help and validation messages in cluster form

### LINKED ISSUE #356: Add popovers to cluster and storage wizards
We need to add popovers to fields like Cluster name, Url, etc. for giving user a hint of what should be there

mturley:
Fixes #356.

Discussing with @vconzola, we decided that to align with [PatternFly design guidelines for forms](https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms), the help text mocked up as popovers in the issue should instead be placed below the text fields using the `helperText` prop of `FormGroup`. Also, this form previously used a special `FormErrorDiv` below the fields to display validation errors, but we should instead use `helperTextInvalid` and mark the `FormGroup`s and `TextInput`s as `isValid={false}` when validation errors are present for that field. This takes up less vertical space, replaces the initial `helperText` if errors are present, and gives the field error styles (red border on the bottom, red alert icon at the right side).

This PR makes these changes in the `AddEditClusterForm` only, but we should use this anywhere we have form help text and validation messages and we should remove `FormErrorDiv` once we've done this everywhere. I opened a new issue for that: https://github.com/fusor/mig-ui/issues/692

This PR also fixes a bug where the `azureResourceGroup` field would not display its validation error due to a typo in the validate function.

# Before

## No validation errors
Screenshot 2020-02-21 16 32 32

## With validation errors
Screenshot 2020-02-21 16 32 52

# After

## No validation errors
Screenshot 2020-02-21 16 23 47

## With validation errors
Screenshot 2020-02-21 16 24 17

### DIFF
--- a/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
+++ b/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
@@ -13,7 +13,6 @@ import {
 import { OutlinedQuestionCircleIcon } from '@patternfly/react-icons';
 import { withFormik, FormikProps } from 'formik';
 import KeyDisplayIcon from '../../../common/components/KeyDisplayIcon';
-import FormErrorDiv from '../../../common/components/FormErrorDiv';
 import HideWrapper from '../../../common/components/HideWrapper';
 import utils from '../../../common/duck/utils';
 import {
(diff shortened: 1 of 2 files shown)


mturley:
Thanks for catching that @ibolton336 , I had left a `||` there instead of `&&`. My first pass had all those `isValid` props defined as `{!touched.foo || !errors.foo}` and I changed it to `{!(touched.foo && errors.foo)} so it would match the `helperTextInvalid` expressions and be a bit more intuitive. Missed that one when making that switch.

Just amended and force-pushed to fix it.

PR #698, opened 2020-02-25
Clean up remaining form validation code, remove FormErrorDiv

### LINKED ISSUE #692: Form validation errors should be displayed using props of PatternFly components instead of FormErrorDiv
See https://github.com/fusor/mig-ui/pull/691. `FormErrorDiv` is also used in `AWSForm`, `AzureForm` and `GCPForm`. We should remove it entirely and instead use `helperTextInvalid` and `isValid` props on `FormGroup` and the `isValid` prop on `TextInput` like I did for the cluster form in #691.

mturley:
Fixes #692 

This PR takes the changes made to the AddEditClusterForm in https://github.com/fusor/mig-ui/pull/691 and repeats them on the rest of the forms that were using FormErrorDiv for validation errors (the 3 forms in the Add replication repository modal).

I also found and fixed two validation bugs in these forms:
* In the GCPForm, the "Required" error would never appear since the other error condition ("The bucket name can be between 3 and 63 characters long") is true in the case of an empty value. I moved the character length condition into an else case so "Required" will appear instead if the field is left blank.
* In the AzureForm, the "Azure credentials" textarea field could not show validation errors because of a typo in the validate function.

# AWSForm

## Before
Screenshot 2020-02-24 19 31 37

## After
Screenshot 2020-02-24 19 35 10

# GCPForm

## Before
Screenshot 2020-02-24 19 49 21

## After
Screenshot 2020-02-24 19 48 54

tJNUdCZ47z

# AzureForm

## Before
Screenshot 2020-02-24 19 32 28

## After
Screenshot 2020-02-24 19 36 57

### DIFF
--- a/src/app/common/components/FormError.module.scss
+++ b/src/app/common/components/FormError.module.scss
@@ -1,6 +1,0 @@
-@import 'src/colors' ;
- 
- .formErrorDiv {
-    margin: 1em 0 0 0;
-    color: $statusRed;
- }
--- a/src/app/common/components/FormErrorDiv.tsx
+++ b/src/app/common/components/FormErrorDiv.tsx
@@ -1,9 +1,0 @@
-import * as React from 'react';
-const styles = require('./FormError.module')
-
-// TODO this should be removed in favor of using the `helperTextInvalid` and `isValid` props of `FormGroup`. 
-const FormErrorDiv = (props) => (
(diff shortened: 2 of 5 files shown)


ibolton336:
Awesome job! Great catch on the validation errors & those look like very clean fixes. Everything checks out locally and on a PROD build.

PR #700, opened 2020-02-25
Increase add/edit timeout length to decrease chance of ui being out of sync

### LINKED ISSUE #699: Update Cluster/Storage functionality is not reflecting accurate information
When editing a cluster/storage object & you change the value of the **_SSL verification required_** checkbox, the updated value isn't shown on the form. You have to click the update button twice for the UI to show the correct info

bug?

ibolton336:


eriknelson:
This isn't a real solution to the problem unfortunately because it happens to rely on the cluster conditions at that particular point in time. I think we need to leave this issue open and fix it deterministically, but we can accept this PR as a band-aid.

PR #703, opened 2020-02-26
Update namespace table values 

### LINKED ISSUE #660: Need to implement the unused columns on the namespace selection screen
image

As seen in the image, we have 3 unpopulated columns still present that still are not hooked up to anything. The first question I'd like to answer with @vconzola and @jwmatthews is, do we still think these columns are relevant and they're desired to keep around? If that's true, we need to enhance the discovery service to provide this information as part of the nameplace discovery endpoint.

I'd also point out these have been empty for a few releases as this point and I haven't heard anything about it. Do we think this columns are really necessary, or would it be useful to do some kind of a UX redesign here built around *only* displaying namespace names.

Related controller issue: https://github.com/fusor/mig-controller/issues/414

ibolton336:
Closes #660 

- Uses newly available namespace data from discovery service api
- Moves namespace fetch function to sagas file from old operators thunk file.

### DIFF
--- a/src/app/plan/components/Wizard/NameSpaceTable.tsx
+++ b/src/app/plan/components/Wizard/NameSpaceTable.tsx
@@ -170,7 +170,7 @@ const NamespaceTable: React.FunctionComponent<INamespaceTableProps> = props => {
                       fontWeight: 600,
                     }}
                   >
-                    Number of Services
+                    Number of services
                     </div>
                 ),
                 accessor: 'serviceCount',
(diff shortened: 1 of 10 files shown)


ibolton336:
Blocked by updates on the controller side. This is tested an working with a prototype controller image

ibolton336:
Screen Shot 2020-02-26 at 4 34 18 PM

@vconzola Look ok to you?

PR #715, opened 2020-03-06
Remove all auth thunks

ibolton336:
Replaced all auth related thunks. Started working on this in an effort to create a patternly compliant way to allow the user to refresh the token via a pop up modal as outlined here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/modal#report-errors

Follow up PR to come with implementation of the error modal with refresh token button. This should provide for a better user experience when a users token expires & cleans up an edge case when the user is not auto rerouted when the session has expired.


eriknelson:
> Follow up PR to come with implementation of the error modal with refresh token button. This should provide for a better user experience when a users token expires & cleans up an edge case when the user is not auto rerouted when the session has expired.

This will be good to have, are there any thunks left after this one? We should shoot for **thunk-zero** (TM) and pull the dep.

ibolton336:
> > Follow up PR to come with implementation of the error modal with refresh token button. This should provide for a better user experience when a users token expires & cleans up an edge case when the user is not auto rerouted when the session has expired.
> 
> This will be good to have, are there any thunks left after this one? We should shoot for **thunk-zero** (TM) and pull the dep. I'd like to test this one in an environment before merging.

Haha there are a few thunks left. They are quite embedded in our app. I am working on doing the plan operations -> thunk conversion right now which will be my next PR. Going to break the conversions up by resource type to make this a little more manageable & testable.

PR #717, opened 2020-03-06
Rename aws to s3 to clarify self hosted storage option

### LINKED ISSUE #669: It is confusing that you must choose AWS type to configure your own self hosted S3
From a slack inquiry:

> Hi all, while adding replication repos to migration tool, I remember it supported any S3 storage (minio in my case). I have a build that only allows provider as aws/gcp/azure. Is that added new or am I using an older version? Thanks.

Right now, when you choose "Add Replication Repository", you are presented a dropdown with three options: "AWS", "Azure", "GCP". If you want to configure your own self hosted S3 bucket (minio or noobaa), you need to actually select AWS. The truth is, the same form is used to configure AWS *or* minio.

@vconzola What is your recommendation here? Should we just update the dropdown to read "S3" instead of AWS?

ibolton336:
Closes #669

### DIFF
--- a/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
+++ b/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
@@ -86,9 +86,9 @@ const AddEditStorageForm = (props: IOtherProps) => {
     } : null
   );
   const [providerOptions, setproviderOptions] = useState([
-    { label: 'aws', value: 'aws' },
-    { label: 'gcp', value: 'gcp' },
-    { label: 'azure', value: 'azure' }
+    { label: 'S3', value: 'aws' },
+    { label: 'GCP', value: 'gcp' },
+    { label: 'Azure', value: 'azure' }
   ]);
 
 


vconzola:
@ibolton336 So what are the dropdown options now? Can you include a screenshot. I'm not sure we want to replace AWS with s3, especially if we still have options for other s3 provider types (e.g., Azure). That will be confusing to users.

ibolton336:
> @ibolton336 So what are the dropdown options now? Can you include a screenshot. I'm not sure we want to replace AWS with s3, especially if we still have options for other s3 provider types (e.g., Azure). That will be confusing to users.

The options are S3, GCP, and Azure. @eriknelson Mentioned that this would be desirable based on the feedback he received.

eriknelson:
To be clear, Azure is not S3, it is their own branded block type storage. S3 is an API standard. Self-hosted options like Minio and Noobaa are both S3, in addition to Amazon S3 itself. So it's correct to have S3, Azure, and Google as storage types.

vconzola:
@eriknelson Cool. Thanks for the explanation.

PR #718, opened 2020-03-06
Set logo pointer to default

### LINKED ISSUE #676: Logo is styled to appear "clickable" but does nothing
Hard to get a screenshot of this, but the logo in the top bar appears to be clickable, but does nothing. @vconzola I think we should decide if this should route to the main page (if, for example, you are at the logs page), or if this style should just be turned off.

ibolton336:
Closes #676

### DIFF
--- a/src/app/common/components/HeaderComponent.module.scss
+++ b/src/app/common/components/HeaderComponent.module.scss
@@ -1,0 +1,3 @@
+.logoPointer {
+    cursor: default !important;
+}
--- a/src/app/common/components/HeaderComponent.tsx
+++ b/src/app/common/components/HeaderComponent.tsx
@@ -4,13 +4,12 @@ import {
   PageHeader,
 } from '@patternfly/react-core';
 import openshiftLogo from './CAM_LOGO.svg';
+const styles = require('./HeaderComponent.module');
 
 const HeaderComponent = (
   <PageHeader
     logo={
-      <React.Fragment>
-        <Brand src={openshiftLogo} alt="OpenShift Logo" />
-      </React.Fragment>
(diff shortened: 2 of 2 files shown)


src/app/common/components/HeaderComponent.module.scss
```diff
@@ -0,0 +1,3 @@
+.logoPointer {
+    cursor: default !important;
```
eriknelson: Not great we need to use `!important` too override these PF styles. Can we file an issue upstream and get to the bottom of why this is necessary?

I think we need to do that before this gets out of control

mturley: On my end when I remove `!important` from this it still gets applied just fine. But otherwise the solution here would be to add more specificity to the selector. The `cursor: pointer` style is being inherited from `.pf-c-page__header-brand-link` so we could use `.pf-c-page__header-brand-link.logoPointer` to be safe.

mturley: Oh, weird... with this style of SCSS import I'm not sure how you can even use external (non-compiled) classes in your selectors.

mturley: Got it, there's a `:global()` syntax for that: https://github.com/konveyor/mig-ui/pull/739

ibolton336: @mturley The !important statements in ResultsStep.module.scss are what started my use of !important on all styles relating to patternfly. The .styledCard & .centerCard styles were working locally but were being overridden in the production image.

ibolton336: css_important

PR #721, opened 2020-03-09
move storage and cluster operations to saga

### LINKED ISSUE #688: Refactor any uses of redux-thunk over to redux saga
Right now in the UI codebase, there are some areas that are particularly hard to follow. Moving to one consistent way of handling CRUD operations or "side effects" would go a long way in terms of code legibility. There are only a few areas of the app still using redux-thunk. 

While this is being done, I think it would make sense to rethink how we have our saga files structured.

ibolton336:
Remove remaining thunks 

Closes #688


eriknelson:
This got kind of mangled with the updates, @ibolton336 if you could take a look at the build failures I think there was some weird merge conflicts limited to a couple places. I did pull it down previously and tested it fine in a cluster, but it just needs some care before it's ready to go. I'll retest once you've had a chance to look.

PR #733, opened 2020-03-10
Remove stage/migrate/plan thunks & improve error handling for stage/migrate.

### LINKED ISSUE #674: Add alerts with detailed message describing why a migration/stage has failed
Via @vconzola 

image

When there is an _error_ condition, currently the error message is shown inline within the data list view. This issue came to light when I tried to add the same functionality for _critical_ conditions, the message was very verbose & took up too much space. We should move this information over to the existing alert system.

### LINKED ISSUE #688: Refactor any uses of redux-thunk over to redux saga

### LINKED ISSUE #719: Handle migration postponed status condition
```status:
  conditions:
  - category: Critical
    lastTransitionTime: "2020-03-06T19:17:01Z"
    message: Postponed 10 seconds to ensure migrations run serially and in order.
    status: "True"
    type: Postponed
  observedGeneration: 40
```

When the critical condition is shown for a postponed migration, the status is updated to failed. This is a regression stemming from https://github.com/konveyor/mig-ui/pull/662/files

ibolton336:
Closes #688 
Closes #719


ibolton336:
updates_stage_mig

ibolton336:
@eriknelson @vconzola There are some issues with only showing failed for all critical conditions. Sometimes, a critical condition means that the migration/stage is postponed for some time. Do we want to add code to handle this special edge case? What should the table data display for this condition? Also, if there is an error it feels like the user could benefit from the error text being displayed continuously rather than within an alert box temporarily. Thoughts on this?

ibolton336:
not_ready

eriknelson:
@vconzola I'm not sure exactly how the conversation has gone wrt these error messages. I think there was an objection to having potentially large inline errors on the migration rows should they error out and instead we wanted to just reported "Failed", but I did have a feedback item from Madrid stating that wasn't useful for them. At the very least, I think we want to re-introduce the error message shown here in the short term.

I reopened #674 for us to have a design discussion about how we want the design to surface stage/migration failure errors in a more design-correct way. Let's start the discussion on that issue so we can get a long term solution in place.

vconzola:
@ibolton336 Re: "Sometimes, a critical condition means that the migration/stage is postponed for some time." If that's the case, it shouldn't be a critical condition. In fact, if it only postpones the migration for 20 seconds there's no reason to even surface it to the user.

eriknelson:
@vconzola there is a lot of unrelated features in this PR that actually had already been merged but we needed to backout to repair regressions. I'd like to move that type of discussion to #674 so we can get this merged asap to capture the tech-debt related refactoring, which is the bulk of this.

PR #740, opened 2020-03-12
Revert "Allow "host" to be selected as source"

ibolton336:
Reverts konveyor/mig-ui#735


ibolton336:
@eriknelson I think we should revert this for now since it is blowing the app up when a user selects host as source cluster.

PR #745, opened 2020-03-13
Patternfly bump

ibolton336:
Need to bump the pf4 version to bring in the file upload component. **_This includes a hacky workaround for an upstream pf4 issue. We are introducing some @ts-ignore statements to be removed at a later date_**

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,8 +62,8 @@
     "webpack-dev-server": "^3.2.1"
   },
   "dependencies": {
-    "@patternfly/react-core": "^3.134.2",
-    "@patternfly/react-table": "^2.25.6",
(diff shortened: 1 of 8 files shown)


eriknelson:
I'm okay merging this as long as we:
* Tag all every usage of the ts ignore so that we can easily revert them when PF has resolved the issue on their side
* File an issue with Patternfly that describes the problem we're seeing since we shouldn't have to include these workarounds
* File an issue in our own repo that links to the Patternfly issue so that once it's fixed in the dependency, we can revert all the workarounds in our repo

ibolton336:
https://github.com/konveyor/mig-ui/issues/747

PR #749, opened 2020-03-13
Replace react-table with PF Table in Resources step of plan wizard

### LINKED ISSUE #585: Migration plan wizard - Step 2 enhancements
There are a number of small UX changes that need to be made to this page, but I want to wait until John has a chance to weigh in on the table contents to do a mockup.

mturley:
Closes #585.

This replaces the `react-table` which was in place in this wizard step with PatternFly Table and Pagination components. In the process, I removed local state and adopted Formik's state as the source of truth for which rows are selected, similarly to #738. This made all of the `useEffect` logic in this file unnecessary.

Screenshot 2020-03-13 19 21 07

While implementing this, I got stuck for a while on an issue with the scope accessible from the `onSelect` function being out-of-date. I think this is a bug in the PF Table component, and I plan to reproduce it in a CodeSandbox and open an issue in PatternFly for that next week. For the time being, I worked around it by passing a reference to current selection state into each row object, and referencing that `rowData` in `onSelect` instead of directly referencing the component's main function scope.

### DIFF
--- a/src/app/plan/components/Wizard/NameSpaceTable.tsx
+++ b/src/app/plan/components/Wizard/NameSpaceTable.tsx
@@ -23,65 +23,12 @@ const NamespaceTable: React.FunctionComponent<INamespaceTableProps> = props => {
 
   if (values.sourceCluster === null) return null;
 
-  /*
-  update formik when local state changes -- no need
   useEffect(() => {
-    if (sourceClusterNamespaces.length > 0) {
-      const formValuesForNamespaces = sourceClusterNamespaces
-        .filter(item => {
-          const keys = Object.keys(checkedNamespaceRows);
+    console.log('>>>>>>>>>>> SELECTED NAMESPACES CHANGED', values.selectedNamespaces);
+  }, [values.selectedNamespaces]);
 
-          for (const key of keys) {
-            if (item.name === key && checkedNamespaceRows[key]) {
-              return item;
-            }
-          }
-        })
-        .map(namespace => namespace.name);
-
-      setFieldValue('selectedNamespaces', formValuesForNamespaces);
-    }
-  }, [checkedNamespaceRows]);
-  */
-
-  /* initialize local state from formik -- no need
-  useEffect(() => {
-    if (values.selectedNamespaces.length > 0 && sourceClusterNamespaces.length > 0) {
-      const newSelected = Object.assign({}, checkedNamespaceRows);
-      values.selectedNamespaces.filter((item, _) => {
-        for (let i = 0; sourceClusterNamespaces.length > i; i++) {
-          if (item === sourceClusterNamespaces[i].name) {
-            newSelected[item] = true;
-          }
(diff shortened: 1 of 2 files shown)


src/app/plan/components/Wizard/NameSpaceTable.tsx
```diff
@@ -1,190 +1,126 @@
...
+    { title: 'Name' },
+    { title: 'Pods' },
+    { title: 'PV claims' },
+    { title: 'Services' },
+  ];
+  const rows = sourceClusterNamespaces.map(namespace => ({
+    cells: [namespace.name, namespace.podCount, namespace.pvcCount, namespace.serviceCount],
+    selected: values.selectedNamespaces.includes(namespace.name),
+    meta: { selectedNamespaces: values.selectedNamespaces }, // See comments on onSelect
+  }));
```
ibolton336: 🏆

eriknelson:
@vconzola this looking good to you?

eriknelson:
Assumed we wanted to get #745 merged before this one, @mturley let me know if this needs any changes after a rebase, otherwise I think we're good to review and merge this :+1:

mturley:
@eriknelson since the whole pane can scroll if the table is tall enough, I don't think we can avoid that. Allowing the popover to overflow would allow the whole table to overflow into the footer with the buttons.

I do know @vconzola wanted to move the table to its own wizard step to give it some more space, which would help with this.

mturley:
Actually, scratch that! We can't make the dropdown pop out of the scroll container, but we can make it "drop up" instead. I'll make that change for the pagination at the bottom of the table.

mturley:
Made that change and rebased on master. How's this @eriknelson ?

Screenshot 2020-03-16 11 56 40

vconzola:
@mturley This looks good to me. One question...when the number of namespaces gets large enough to require scrolling, will the table scroll or the page scroll?

mturley:
@vconzola the whole body of the wizard step will scroll (not including the footer with the buttons).

eriknelson:
@vconzola Are we okay with that scrolling behavior? Under most circumstances, this is giong to be the behavior given an average namespace count.

mturley:
@vconzola here's how it looks (including mobile responsiveness):
cV3VjSQgVo

eriknelson:
Ah sorry, my concern was if you had something like 100 namespaces, is it paged? Sounded like it would be a long single scrollable table unless I misunderstood? 

⁣Get BlueMail for Android ​

mturley:
Oh yeah sorry for the confusion there @eriknelson, the table is paged (defaults to 10 namespaces per page, can be changed up to 100 per page). It's just that the current page of the table is part of the scroll container that also includes the dropdowns at the top. So you'll still have to scroll to get to the bottom of the current page.

vconzola:
Thanks @mturley. This looks good.

PR #750, opened 2020-03-14
Deprecate unused integration tests

eriknelson:
* Remove cypress dependency
* Update yarn.lock
* Archive old files

~Working through fixing travis, this should fail until updated.~

Travis is functioning as it should be; it's failing bad lints, running lone tests, and at least verifying builds. There's a lot hanging around the project that's legacy and generally hasn't been maintained. I'd like to either clean it and get rid of it, or update it so it's actually useful. Also, we need to have a broader conversation about:

* What should we do about the mocked environment, because it's not functional at all right now.
* What exactly is our test strategy going forwards?


ibolton336:
Just need to fix that yarn.lock conflict otherwise LGTM

PR #757, opened 2020-03-19
Modal to confirm plan deletion

gildub:
confirm-delete-plan

Closes https://github.com/konveyor/mig-ui/issues/729

### DIFF
--- a/src/app/home/components/DataList/Plans/PlanActions.tsx
+++ b/src/app/home/components/DataList/Plans/PlanActions.tsx
@@ -13,9 +13,11 @@ import { useOpenModal } from '../../../duck/hooks';
 import MigrateModal from '../../../../plan/components/MigrateModal';
 import { withRouter } from 'react-router-dom';
 import WizardContainer from '../../../../plan/components/Wizard/WizardContainer';
+import ConfirmModal from '../../../../common/components/ConfirmModal';
 
 const PlanActions = ({ plan, history }) => {
   const [isMigrateModalOpen, toggleMigrateModalOpen] = useOpenModal(false);
+  const [isDeleteModalOpen, toggleDeleteModalOpen] = useOpenModal(false);
   const [isWizardOpen, toggleWizardOpen] = useOpenModal(false);
   const planContext = useContext(PlanContext);
   const {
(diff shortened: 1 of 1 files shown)


eriknelson:
Thanks @gildub. Anything that is making user facing changes like this, we need to ping @vconzola to get his ACK prior to a merge. Could you also add the UXD label on the PR In the future? Thanks!

vconzola:
@gildub Buttons should be side by side (with Confirm on the left) rather than stacked. Spacing between the text and the buttons should be 16px. Thanks.

gildub:
@vconzola, please review, buttons are now side by side.

vconzola:
@gildub Thanks. Buttons and spacing look perfect. One question...did you change the modal heading? Didn't it used to say "Confirm delete" or "Confirm deletion"? I liked that better.

gildub:
@vconzola, yes I changed the modal by reusing the existing component "ConfirmModal", instead of creating a component for it. This is simpler but the trade off is the lack of header besides "Confirmation" . Let me see if we can make ConfirmModal more generic.

ibolton336:
@gildub Can you refrain from rebasing? and squashing commits? It makes it harder to follow the requested changes & your solution. Sorry we should add that to the CONTRIBUTING.md & maybe see if we can enforce disabling force push @eriknelson .. that might keep us all honest.

eriknelson:
We can't disable force push to forks which should be the normal way code
changes are brought in via PR, but you're right we should add that to the
contributing doc. The repo is set up in a way where PRs are automatically
squashed before they are brought into master. This keeps master's commit
log clean on a PR basis, but an individual PR can remain a series of
commits that tell the story of a particular PR. Abstaining from squashing
and force pushing helps reviewers because they can review commits
independent from the broader PR. Makes it much easier to ACK after
requested changes, in particular.

gildub:
@ibolton336, @eriknelson, I'll definitely avoid force-pushing unless this is early stage of a PR such as [WIP] and not being reviewed yet. I agree that it makes the review more difficult.

mturley:
@ibolton336 @gildub this PR causes a bug, `planContext.handleDeletePlan(plan)` is no longer called anywhere, so clicking the confirm button on this modal does not actually delete the plan

PR #759, opened 2020-03-20
MigCluster : Use generateName to create unique secrets

### LINKED ISSUE #396: Clusters and Storage sharing the same name will have secrets fail to create

pranavgaikwad:
Fixes #396 

Proposed Changes :
[1] Creating new MigCluster 
- We use `generateName` instead of `name` in tokenSecret of MigCluster. We also introduce two new labels on the Secret : `createdForResourceType`, `createdForResource`. This makes it easier to find a secret associated with a cluster.
- We create secret first, then get the resulting secret name, then we create MigCluster

[2] Deleting MigCluster
- We find the secret name before deleting MigCluster
- Then usual flow of delete

[3] Editing MigCluster
- Nothing changes here except the secret reference.

### DIFF
--- a/src/app/cluster/duck/sagas.ts
+++ b/src/app/cluster/duck/sagas.ts
@@ -89,9 +89,11 @@ function* removeClusterSaga(action) {
     );
     const migClusterResource = new MigResource(MigResourceKind.MigCluster, migMeta.namespace);
     
-    const secretResourceList = yield client.list(secretResource, getTokenSecretLabelSelector(MigResourceKind.MigCluster, name))
+    const secretResourceList = yield client.list(secretResource, 
+      getTokenSecretLabelSelector(MigResourceKind.MigCluster, name));
 
-    const secretResourceName = secretResourceList.data.items[0].metadata.name;
+    const secretResourceName = (secretResourceList.data.items && secretResourceList.data.items.length > 0) 
+      ? secretResourceList.data.items[0].metadata.name : '';
 
     yield Promise.all([
       client.delete(secretResource, secretResourceName),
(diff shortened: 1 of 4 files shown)


src/app/cluster/duck/sagas.ts
```diff
@@ -304,7 +327,7 @@ function* updateClusterRequest(action) {
 
     // Pushing a request fn to delay the call until its yielded in a batch at same time
     updatePromises.push(() => client.patch(
-      secretResource, clusterValues.name, newTokenSecret));
+      secretResource, currentCluster.spec.serviceAccountSecretRef.name, newTokenSecret));
```
pranavgaikwad: Here I am assuming that Cluster name is not changeable from UI

pranavgaikwad:
@ibolton336 added a fix to address the problem you saw. safe guarded the case when - associated Secret does not exist and we try to delete a cluster.

src/app/cluster/duck/sagas.ts
```diff
@@ -132,13 +142,17 @@ function* addClusterRequest(action) {
...
     ]);
 
     const alreadyExists = getResults.reduce((exists, res) => {
-      return res.value && res.value.status === 200 ?
-        [...exists, { kind: res.value.data.kind, name: res.value.data.metadata.name }] :
+      return res && res.status === 200 ?
+        [...exists, { kind: res.value.data.kind, 
+          name:  (res.value.data.items && res.value.data.items.length > 0) 
+            ? res.value.data.items[0].metadata.name : res.value.data.metadata.name
+        }] :
```
ibolton336:
```suggestion
      return (res && res.status === 200) ?
        [...exists,
          {
            kind: res.value.data.kind,
            name: (res.value.data.items && res.value.data.items.length > 0) ?
              res.value.data.items[0].metadata.name : res.value.data.metadata.name
          }
        ] :
```

src/app/cluster/duck/sagas.ts
```diff
@@ -87,9 +88,15 @@ function* removeClusterSaga(action) {
       migMeta.configNamespace
     );
     const migClusterResource = new MigResource(MigResourceKind.MigCluster, migMeta.namespace);
+    
+    const secretResourceList = yield client.list(secretResource, 
+      getTokenSecretLabelSelector(MigResourceKind.MigCluster, name));
+
+    const secretResourceName = (secretResourceList.data.items && secretResourceList.data.items.length > 0) 
+      ? secretResourceList.data.items[0].metadata.name : '';
```
ibolton336:
```suggestion
    const secretResourceName = (secretResourceList.data.items && secretResourceList.data.items.length > 0) ?
      secretResourceList.data.items[0].metadata.name : '';
```

PR #761, opened 2020-03-20
ConfimModal: Add title props

gildub:
Add a title prop to ConfirmModal common component.
Makes use of it to follow PatternFly best practice, for cluster and storage deletion.

confirm-delete-cluster

confirm-delete-replication

### DIFF
--- a/src/app/common/components/ConfirmModal.tsx
+++ b/src/app/common/components/ConfirmModal.tsx
@@ -4,16 +4,17 @@ import { Button, Modal } from '@patternfly/react-core';
 interface IProps {
   onHandleClose: (isConfirmed) => void;
   id: string;
+  title: string;
   message: string;
   isOpen: boolean;
 }
 
-const ConfirmModal: React.FunctionComponent<IProps> = ({ isOpen, onHandleClose, message, id }) => {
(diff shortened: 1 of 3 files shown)


ibolton336:
@vconzola Do we have your blessing of just using "Confirmation" for the title? @gildub Can you provide screenshots of each scenario?

gildub:
@vconzola, what about adding text to "Confirmation" respectively for cluster and replication repository deletion". Maybe "Cluster Deletion Confirmation" and "Replication Repository Deletion Confirmation"?

vconzola:
@gildub @ibolton336 Guidance from PatternFly content strategist is as follows:
The best practice is to have:
A descriptive heading that asks a question 
A couple sentences or so of explanation 
Buttons to answer the question. 

Example:
Delete this report?
Your information won’t be saved.
Yes/No (button text certainly doesn’t have to be yes/no—it just has to answer the heading question)

So in our case it would be:
Remove this cluster?
Removing "clustername" will make it unavailable for migration plans.
Yes/No

gildub:
@vconzola, I updated the screenshots with latest update, please let me know what you think.

gildub:
@vconzola, I removed the capitalization in the heading.

PR #764, opened 2020-03-20
(bug fix): Set correct value types for disable check connection check

### LINKED ISSUE #762: (bug): Check connection button always disabled
https://bugzilla.redhat.com/show_bug.cgi?id=1779164 

When adding new fields to the cluster form, this disable function was not updated properly. n

image

ibolton336:
Closes #762

### DIFF
--- a/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
+++ b/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
@@ -45,18 +45,32 @@ const valuesHaveUpdate = (values, currentCluster) => {
   if (!currentCluster || currentCluster.MigCluster.spec.isHostCluster) {
     return true;
   }
-
   const requireSSL = !currentCluster.MigCluster.spec.insecure;
-  const caBundle = currentCluster.MigCluster.spec.caBundle;
   const rawToken = atob(currentCluster.Secret.data.saToken);
   const existingEndpoint = currentCluster.MigCluster.spec.url;
-  const azureResourceGroup = currentCluster.MigCluster.spec.azureResourceGroup;
+
+  const getAzureResourceGroupVal = (azureResourceGroup) => {
+    if (azureResourceGroup) {
+      return azureResourceGroup.length > 0
+    } else return '';
+  }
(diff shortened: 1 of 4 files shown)


src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
```diff
@@ -45,18 +45,38 @@ const valuesHaveUpdate = (values, currentCluster) => {
...
     return true;
   }
-
   const requireSSL = !currentCluster.MigCluster.spec.insecure;
-  const caBundle = currentCluster.MigCluster.spec.caBundle;
   const rawToken = atob(currentCluster.Secret.data.saToken);
   const existingEndpoint = currentCluster.MigCluster.spec.url;
-  const azureResourceGroup = currentCluster.MigCluster.spec.azureResourceGroup;
+
+  const getAzureResourceGroupVal = (azureResourceGroup) => {
```
eriknelson: This looks dangerous. One branch returns a bool, the other returns a string. Should return something of the same type.

src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
```diff
@@ -45,18 +45,38 @@ const valuesHaveUpdate = (values, currentCluster) => {
...
+    if (azureResourceGroup) {
+      return azureResourceGroup.length > 0
+    } else return '';
```
eriknelson: This is a weird `if...else` format. Can we use something more traditional, with a ternary operator or a standard if/else?

`return azureResourceGroup ? azureResourceGroup.lengh > 0 : '';` (Still need to fix the return type here).

Does the linter like this? We should see if we can update.

ibolton336: +1 updated to ternaries

src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
ibolton336: For isAzure, the controller is expecting a boolean. The azureResourceGroup is a text input. Not sure if i am understanding the issue here?

PR #768, opened 2020-03-24
Replication Repository : delete secret on repo deletion

### LINKED ISSUE #723: Replication Repo delete does not delete related secret
Being careful not to say that the secret is leaked, but a delete of the replication repository does not delete the corresponding secret in the ref (usually in openshift-migration).

pranavgaikwad:
Fixes #723 

Fixed wrong reference to Secret namespace in removeStorageSaga()

We create secret in `migMeta.configNamespace` not `migMeta.namespace`


eriknelson:
+1, nice catch.

Issue #426, opened 2019-08-02
Support source clusters with 10k-100k namespaces


jwmatthews:
We expect source clusters in production will have large numbers of namespaces, estimates of 10k-100k namespaces.

This issue is intended to track:

- Create a script to populate X number of namespaces that are empty
- Load a few sample apps in a few namespaces
- Use the WebUI to perform a migration selecting a few namespaces with apps

Assume to do the above we will see gaps and file new issues.
Assume probably will need support for:

- Support pagination of fetching namespaces from source cluster and iterating
- Filter the returned data to basic on search/filter parameters

PR #771, opened 2020-03-24
[WIP] Add Plan Wizard: Split resources into src, dst and replication

gildub:
WIP - Please DO NOT MERGE.

animation

Closes https://github.com/konveyor/mig-ui/issues/724
Closes https://github.com/konveyor/mig-ui/issues/765

### DIFF
--- a/src/app/plan/components/Wizard/DestinationSelectForm.tsx
+++ b/src/app/plan/components/Wizard/DestinationSelectForm.tsx
@@ -26,7 +26,7 @@ const DestinationSelectForm = props => {
 
   useEffect(() => {
     // ***
-    // * Populate src and target cluster dropdowns
+    // * Populate target cluster dropdown
     // ***
     if (clusterList.length) {
       const targetOptions = [];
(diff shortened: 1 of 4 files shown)


gildub:
@vconzola, please let me know what you think.

mturley:
@vconzola your other reasons to hold off are valid, and @gildub we should communicate better so we don't step on each other's toes, but luckily I think this would not conflict with the stuff I'm working on for the PV and Storage Class wizard steps, since we're working in different files.

eriknelson:
>  I'd also like to make these changes at the same time we're addressing changes to be able to scale the UI to handle large numbers of namespaces #426.

@vconzola It's unlikely we're going to touch #426 until 1.3, certainly it's post-1.2.0.

Are we sure we want to wait until we're moving into June to split out this screen?

vconzola:
@eriknelson Yes, I'd prefer to wait. As implement it would add two steps to the wizard, which with the hooks would mean we'd have an 8 step wizard and three of the first four steps would have only a single input field. I'd rather take some time and understand what fields can be combined in a step in way that makes sense from both a UX and controller standpoint. Also, the wizard works as is and we don't need to break the steps out until we tackle #426, so I think we should do them together.

eriknelson:
@vconzola ACK, makes sense.

@gildub Going to close this PR; I don't think there's going to be a lot here that can be reused and we're going to take a step back from a design standpoint and revisit this in a couple months.

PR #776, opened 2020-03-26
Enhance PVs step of plan wizard: switch to PF table, fix useEffect logic, add types, other UX cleanup

mturley:
Closes https://github.com/konveyor/mig-ui/issues/586

* Lifts non-table-presentation-related logic out of `VolumesTable` into `VolumesForm`.
* Moves table header text down into `VolumesTable` so it doesn't appear during discovery/polling.
* Fixes an issue with the useEffect logic where `getPVResourcesRequest` was being called four times instead of once, causing ~500ms / 2KB of extra network requests (see useEffect note below)
* Converts the PVs table to a PatternFly Table
* Converts the `react-select` components to PatternFly Select
* Abstracts out a new reusable React hook: `usePaginationState` (from code I wrote for `NameSpacesTable` pagination in https://github.com/konveyor/mig-ui/pull/749)
* Adds TypeScript prop type interfaces to `VolumesTable` and `VolumesForm` components, by using `Pick<>` to share types with `WizardContainer`, where I added some more specific types.
* ﻿﻿Changes navigation and table column headers to sentence-case capitalization.

# Screens

## Before:
* Non-PF react-table and react-select
* Header text appears before table is loaded
* Title-case capitalization on nav and headers

Screenshot 2020-04-01 15 54 58

Screenshot 2020-04-01 15 55 21

## After

Screenshot 2020-04-01 15 47 25

Screenshot 2020-04-01 15 47 35

# Note about useEffect

Because of how `useEffect`'s dependency array argument works, it's easy to have `useEffect` logic called more often than necessary. It's important to specify very specific things in that array so that the effect will only run again if that specific thing changes.

In this case, the `useEffect` that waits for discovered PVs to be present and then calls `getPVResourcesRequest` was using `[currentPlan, isPollingStatus]` as its dependencies, which means it was running again every time the component re-rendered (since `currentPlan` comes from redux state, and each state change creates a new immutable `currentPlan`, it is always changing). What we really needed was for that logic to only run when `currentPlan.spec.persistentVolumes` goes from empty to non-empty, so the real dependency is just the length of that array.

On entering the wizard step, network traffic before 2ecb64f:
Screenshot 2020-03-25 19 49 32-bad

Network traffic after 2ecb64f:
Screenshot 2020-03-25 19 53 10-good

### DIFF
--- a/src/app/plan/components/Wizard/VolumesForm.tsx
+++ b/src/app/plan/components/Wizard/VolumesForm.tsx
@@ -53,9 +53,10 @@ const VolumesForm = props => {
     }
   }, []);
 
+  const discoveredPersistentVolumes = (currentPlan && currentPlan.spec.persistentVolumes) || [];
+
   useEffect(() => {
-    if (currentPlan) {
-      const discoveredPersistentVolumes = currentPlan.spec.persistentVolumes || [];
+    if (discoveredPersistentVolumes.length > 0) {
       getPVResourcesRequest(discoveredPersistentVolumes, values.sourceCluster || '');
       let mappedPVs;
       if (values.persistentVolumes) {
@@ -100,7 +101,7 @@ const VolumesForm = props => {
       setFieldValue('persistentVolumes', mappedPVs);
       setRows(mappedPVs); // ???
     }
-  }, [currentPlan, isPollingStatus]); // Only re-run the effect if fetching value changes
+  }, [discoveredPersistentVolumes.length]); // Only re-run the effect if fetching value changes
 
   //TODO: added this component level error state to handle the case of no PVs
   // showing up after 3 checks of the interval. When the isPVError flag is checked,
--- a/src/app/plan/components/Wizard/VolumesForm.tsx
+++ b/src/app/plan/components/Wizard/VolumesForm.tsx
@@ -60,7 +60,7 @@ const VolumesForm = props => {
           }
           // TODO do we really need to generate these meta-objects and keep them in formik?
           // currentPlan.spec.persistentVolumes is in global state, so we could just
-          // infer from that everywhere and keep only a mapping of action selections in formik
+          // infer from that everywhere and keep only a mapping of pv id => type selections in formik
           return {
             name: planVolume.name,
             project: planVolume.pvc.namespace,
@@ -148,7 +148,7 @@ const VolumesForm = props => {
       isEdit={isEdit}
       pvResourceList={pvResourceList}
       isFetchingPVResources={isFetchingPVResources}
-      rows={values.persistentVolumes}
+      persistentVolumes={values.persistentVolumes}
       onTypeChange={onTypeChange}
     />
   );
--- a/src/app/plan/components/Wizard/VolumesForm.tsx
+++ b/src/app/plan/components/Wizard/VolumesForm.tsx
@@ -28,20 +28,6 @@ const VolumesForm = props => {
     isPollingStatus,
   } = props;
 
-  const [rows, setRows] = useState([]); // TODO do we need state here? can we use redux+formik?
-
-  const updateTableData = (rowIndex, updatedValue) => {
-    const rowsCopy = [...rows];
-    if (currentPlan !== null && values.persistentVolumes) {
-      const updatedRow = { ...rowsCopy[rowIndex], type: updatedValue };
-
-      rowsCopy[rowIndex] = updatedRow;
-    }
-
-    setRows(rowsCopy);
-    setFieldValue('persistentVolumes', rowsCopy);
-  };
-
   useEffect(() => {
     //kick off pv discovery once volumes form is reached with current selected namespaces
     let isRerunPVDiscovery = null;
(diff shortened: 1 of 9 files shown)


mturley:
@vconzola you got it. Thanks for catching that.

mturley:
Actually @vconzola the wizard step title is missing from all steps of the wizard, so I think if it's ok with you guys I'm going to do that in a separate PR so this one stays relatively localized to this step. Broken out into a new issue: https://github.com/konveyor/mig-ui/issues/795

src/app/plan/components/Wizard/VolumesForm.tsx
```diff
@@ -34,30 +52,119 @@ const VolumesForm = props => {
...
+            if (rowVal && rowVal.selection) {
+              pvAction = rowVal.selection.action;
+            } else {
+              if (rowVal && rowVal.type) {
+                pvAction = rowVal.type;
+              }
+            }
+          }
+          // TODO do we really need to generate these meta-objects and keep them in formik?
+          // currentPlan.spec.persistentVolumes is in global state, so we could just
```
ibolton336, line 75: +1 Sounds like a good idea to refactor to use currentPlan as the source of truth

src/app/plan/components/Wizard/VolumesForm.tsx
```diff
@@ -34,30 +52,119 @@ const VolumesForm = props => {
...
+            claim: planVolume.pvc.name,
+            type: pvAction,
+            details: '',
+            supportedActions: planVolume.supported.actions,
+          };
+        });
+      }
+      setFieldValue('persistentVolumes', mappedPVs);
+    }
+  }, [discoveredPersistentVolumes.length]); // Only re-run the effect if fetching value changes
```
ibolton336, line 105: +1 Nice job with the useEffect enhancements. Originally we were using useEffect to filter out all of the polling updates in the background causing rerenders. Glad this is getting some attention! Looks much better.

mturley: Ah that makes sense. Unnecessary re-renders are still a thing I never know how to handle correctly.

src/app/plan/components/Wizard/WizardContainer.tsx
```diff
@@ -26,19 +31,22 @@ export interface IOtherProps {
...
   stopPlanStatusPolling: (planName) => void;
   pvUpdatePollStop: () => void;
   planUpdateRequest: (values, isRerunPVDiscovery) => void;
   resetCurrentPlan: () => void;
   setCurrentPlan: (plan) => void;
   fetchNamespacesRequest: (clusterName) => void;
-  getPVResourcesRequest: () => void;
+  getPVResourcesRequest: (
+    persistentVolumes: IPlanPersistentVolume[],
+    sourceClusterName: IFormValues['sourceCluster']
```
ibolton336, line 45: image

^^^ so satisfying! Nice job building out the typescript support here.

src/app/plan/components/Wizard/types.ts
```diff
@@ -0,0 +1,35 @@
+// TODO add other properties used by components in this wizard, possibly move this somewhere more common
+// Maybe in the future we can somehow generate types like these from the CRDs in mig-controller?
+
+export interface IPlanPersistentVolume {
```
ibolton336, line 4: +1 i am on board with the types.ts precedent for adding types to components going forward for now.

src/app/plan/components/Wizard/VolumesTable.tsx
```diff
@@ -16,316 +12,143 @@ import {
...
-            <Spinner size="xl" />
-          </div>
-          <Title headingLevel="h2" size="xl">
-            Discovering persistent volumes attached to source projects...
-          </Title>
-        </EmptyState>
-      </Bullseye>
+  const rows = persistentVolumes.map((pv, pvIndex) => {
+    const matchingPVResource = pvResourceList.find(pvResource => pvResource.name === pv.name);
+    const migrationTypeOptions = pv.supportedActions.map(
```
ibolton336, line 66: seeing the app blow up here in some cases. I think we need a guard if supportedActions is undefined. 

image

ibolton336: image

ibolton336: Steps to reproduce: navigate back after selecting namespaces, select a new namespace & deselect old namespace, try to navigate forward and run pv discovery.

mturley: Ah interesting... I wonder why you're getting the original PV from redux there and not the meta-PV-thingy defined here: https://github.com/konveyor/mig-ui/pull/776/files#diff-342db03f4d82174c47390c429f6aa190R77-R85 ... I'll take a look.

mturley: It's strange that the same error wasn't being caused by this: https://github.com/konveyor/mig-ui/pull/776/files#diff-8d2e158fee7fbe8447115eb6c8a9725eL260-L264

mturley: Wrote up an issue about this and the refactor I reference in a few of these comments: https://github.com/konveyor/mig-ui/issues/797

PR #780, opened 2020-03-27
Wizard: Destructuring props directly and use TypeScript types along with types annotations

gildub:
Destructuring props directly in component instead of using proxy props object.
This allows props validation and to make sure unnecessary properties aren't passed to components

### DIFF
--- a/src/app/plan/components/Wizard/GeneralForm.tsx
+++ b/src/app/plan/components/Wizard/GeneralForm.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import { RouteComponentProps } from 'react-router-dom';
+import { FormikProps, FormikActions, FormikState } from 'formik';
+import { IFormValues, IOtherProps } from './WizardContainer';
 import {
   Form,
   FormGroup,
@@ -11,15 +13,25 @@ interface IProps {
   component: React.ReactNode;
 }
 
-const GeneralForm: React.SFC<IProps & RouteComponentProps> = ({
-  handleChange,
-  handleBlur,
-  values,
+interface IGeneralFormProps
+  extends Pick<
+    IOtherProps,
+    | 'isEdit'
+    >,
+    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
+    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
+    Pick<FormikState<IFormValues>, 'touched'> {
+      errors: any;
+    }
+const GeneralForm: React.FunctionComponent<IGeneralFormProps & RouteComponentProps> = ({
   errors,
-  touched,
+  handleBlur,
+  handleChange,
+  isEdit,
   setFieldTouched,
-  isEdit
-}) => {
+  touched,
+  values,
+}: IGeneralFormProps & RouteComponentProps) => {
   const onHandleChange = (val, e) => {
     handleChange(e);
   };
--- a/src/app/plan/components/Wizard/GeneralForm.tsx
+++ b/src/app/plan/components/Wizard/GeneralForm.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { RouteComponentProps } from 'react-router-dom';
 import { FormikProps, FormikActions, FormikState } from 'formik';
 import { IFormValues, IOtherProps } from './WizardContainer';
 import {
@@ -14,13 +13,21 @@ interface IProps {
 }
 
 interface IGeneralFormProps
-  extends Pick<IOtherProps, 'isEdit'>,
-    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
-    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
-    Pick<FormikState<IFormValues>, 'touched'> {
+  extends Pick<
+    IOtherProps, 'isEdit'
+    >,
+    Pick<FormikProps<IFormValues>, 
+    | 'handleBlur'
+    | 'handleChange'
+    | 'setFieldTouched' 
+    | 'values' 
+    | 'touched'
+    > 
+    {
       errors: any;
     }
-const GeneralForm: React.FunctionComponent<IGeneralFormProps & RouteComponentProps> = ({
+
+const GeneralForm: React.FunctionComponent<IGeneralFormProps> = ({
   errors,
   handleBlur,
   handleChange,
@@ -28,7 +35,7 @@ const GeneralForm: React.FunctionComponent<IGeneralFormProps & RouteComponentPro
   setFieldTouched,
   touched,
   values,
-}: IGeneralFormProps & RouteComponentProps) => {
+}: IGeneralFormProps) => {
   const onHandleChange = (val, e) => {
     handleChange(e);
   };
--- a/src/app/plan/components/Wizard/GeneralForm.tsx
+++ b/src/app/plan/components/Wizard/GeneralForm.tsx
@@ -14,10 +14,7 @@ interface IProps {
 }
 
 interface IGeneralFormProps
-  extends Pick<
-    IOtherProps,
-    | 'isEdit'
-    >,
+  extends Pick<IOtherProps, 'isEdit'>,
     Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
     Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
     Pick<FormikState<IFormValues>, 'touched'> {
--- a/src/app/plan/components/Wizard/ResourceSelectForm.tsx
+++ b/src/app/plan/components/Wizard/ResourceSelectForm.tsx
@@ -18,11 +18,11 @@ interface IResourceSelectFormProps
   extends Pick<
     IOtherProps,
     | 'clusterList'
-    | 'storageList'
-    | 'isFetchingNamespaceList'
     | 'fetchNamespacesRequest'
-    | 'sourceClusterNamespaces'
     | 'isEdit'
+    | 'isFetchingNamespaceList'
+    | 'sourceClusterNamespaces'
+    | 'storageList'
     >,
     Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
     Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
@@ -32,16 +32,16 @@ interface IResourceSelectFormProps
 
 const ResourceSelectForm: React.FunctionComponent<IResourceSelectFormProps> = ({
   clusterList,
-  storageList,
-  values,
   errors,
-  touched,
-  setFieldValue,
-  setFieldTouched,
-  isFetchingNamespaceList,
   fetchNamespacesRequest,
-  sourceClusterNamespaces,
   isEdit,
+  isFetchingNamespaceList,
+  setFieldTouched,
+  setFieldValue,
+  sourceClusterNamespaces,
+  storageList,
+  touched,
+  values,
 }: IResourceSelectFormProps) => {
   useEffect(() => {
     if (isEdit) {
--- a/src/app/plan/components/Wizard/ResourceSelectForm.tsx
+++ b/src/app/plan/components/Wizard/ResourceSelectForm.tsx
@@ -24,9 +24,13 @@ interface IResourceSelectFormProps
     | 'sourceClusterNamespaces'
     | 'storageList'
     >,
-    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
-    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
-    Pick<FormikState<IFormValues>, 'touched'> {
+    Pick<FormikProps<IFormValues>, 
+    | 'setFieldTouched'
+    | 'setFieldValue' 
+    | 'touched'
+    | 'values'
+    > 
+    {
       errors: any;
     }
 
--- a/src/app/plan/components/Wizard/ResourceSelectForm.tsx
+++ b/src/app/plan/components/Wizard/ResourceSelectForm.tsx
@@ -1,4 +1,6 @@
 import React, { useState, useEffect } from 'react';
+import { FormikProps, FormikActions, FormikState } from 'formik';
+import { IFormValues, IOtherProps } from './WizardContainer';
 import {
   Bullseye,
   EmptyState,
@@ -12,21 +14,35 @@ import NamespaceTable from './NameSpaceTable';
 import { Spinner } from '@patternfly/react-core/dist/esm/experimental';
 import SimpleSelect from '../../../common/components/SimpleSelect';
 
-const ResourceSelectForm = props => {  
-  const {
-    clusterList,
-    storageList,
-    values,
-    errors,
-    touched,
-    setFieldValue,
-    setFieldTouched,
-    isFetchingNamespaceList,
-    fetchNamespacesRequest,
-    sourceClusterNamespaces,
-    isEdit,
-  } = props;
+interface IResourceSelectFormProps
+  extends Pick<
+    IOtherProps,
+    | 'clusterList'
+    | 'storageList'
+    | 'isFetchingNamespaceList'
+    | 'fetchNamespacesRequest'
+    | 'sourceClusterNamespaces'
+    | 'isEdit'
+    >,
+    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
+    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
+    Pick<FormikState<IFormValues>, 'touched'> {
+      errors: any;
+    }
 
+const ResourceSelectForm: React.FunctionComponent<IResourceSelectFormProps> = ({
+  clusterList,
+  storageList,
+  values,
+  errors,
+  touched,
+  setFieldValue,
+  setFieldTouched,
+  isFetchingNamespaceList,
+  fetchNamespacesRequest,
+  sourceClusterNamespaces,
+  isEdit,
+}: IResourceSelectFormProps) => {
   useEffect(() => {
     if (isEdit) {
       fetchNamespacesRequest(values.sourceCluster);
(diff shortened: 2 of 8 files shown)


src/app/plan/components/Wizard/ResourceSelectForm.tsx
```diff
@@ -88,6 +87,7 @@ const ResourceSelectForm = props => {
     const matchingCluster = clusterList.find(c => c.MigCluster.metadata.name === value);
     if (matchingCluster) {
       setFieldValue('sourceCluster', value);
+      setFieldValue('selectedNamespaces', []);
```
ibolton336: Should seperate this into its own PR for testing purposes

src/app/plan/components/Wizard/StorageClassTable.tsx
```diff
@@ -119,7 +124,7 @@ const StorageClassTable = (props) => {
 
   if (rows !== null) {
     return (
-      <React.Fragment>
+      <>
```
ibolton336: LGTM - but would like @mturley & @eriknelson opinion on switching to this syntax just so we can all agree this would be a worthwhile change. Only downside I see is if you wanted to add a key to the React.Fragment but I dont think we are doing that anywhere.

mturley: I like the shorthand, so LGTM but I don't have a strong opinion. In the case of needing a key we can always just use React.Fragment for those specific fragments.

src/app/plan/components/Wizard/ResourceSelectForm.tsx
gildub: @ibolton336, right, that's a mistake, thanks.

mturley:
@gildub I'm all for this as a convenience/style change, but I'm surprised it would be necessary to catch props being overloaded or missing. Doesn't TypeScript catch that for us in places where we improperly use each component, because of the `React.FunctionComponent<INamespaceTableProps>` syntax?

Also, what was the reason for changing e.g. `INamespaceTableProps` to `INamespaceTable`? I find it is generally useful to have "Props" in those interface names to reason about the fact that it is a list of prop types, if you need to do things like have one component's props extend another's props. `INamespaceTable` looks almost identical to `NamespaceTable`, if we'd change anything there I would just drop the I and call it `NamespaceTableProps` etc. Matter of personal preference though, I'm just curious why you made the decision.

eriknelson:
> Doesn't TypeScript catch that for us in places where we improperly use each component, because of the React.FunctionComponent<INamespaceTableProps> syntax?

+1. We weren't very disciplined about this from the beginning and as we were learning TS. Ideally we have a props interface defined for all our components so bad prop usage doesn't even compile and is caught at authorship time.

> I find it is generally useful to have "Props" in those interface names to reason about the fact that it is a list of prop types, if you need to do things like have one component's props extend another's props. 

+1.

mturley:
@eriknelson I bet there is a `@typescript-eslint` rule we can use to require every component to have its prop types specified. Can't find it at the moment, but it might be a nice thing to enforce later on.

eriknelson:
> Destructuring separately from the function parameters doesn't catch properties overload or missisng ones either, see respectively:

Not sure what's meant by property overload?

Missing properties are valid as they can be optional, and should be marked as such in the prop interface.

gildub:
@mturley, my approach with props is that we shouldn't have to deal with the props name itself unless there is no other choice. From there is like dubbing a type Foo FooType. So wiith such train of thoughts I changed the type too.

eriknelson:
> From there is like dubbing a type Foo FooType.

I don't agree with this. The `Props` suffix is on a very limited amount of interfaces and indicates a well-defined semantic purpose within the context in a way `Type` does not. Including that as part of the name makes it very easy to identify exactly what the intention is for that interface just by reading the name.

gildub:
@eriknelson, you're right, the props has nothing to do with the type and using props in the name of the interface definition is not like doing FooType. Related to that would be more the I which I guess is for Interface. 

`interface INamespaceTableProps { ...}`

The idea is to avoid using with declared names the technical structure involved in the object.

gildub:
@eriknelson, by overload I meant that the sender is passing props not needed by component.
Missing properties are not optional when calling component directly as in the changed code.
This is actually how I discovered some were xtra. 
The patch might seems like changing a lot of things, but it's mostly because I have ordered the props (from senders too) to make it clear (and trying to be consistent with what I've seen in some places).

mturley:
> passing props not needed by component

I would think this would already have been caught by TypeScript just by those props not being part of the props interface, so passing extra ones should have already given us a "X cannot be assigned to Y" error. I don't see how destructuring the props directly in the argument list changes anything about how the types are checked, what error can we get with that that we can't get without it? (Again I'm not opposed to the change, I like that destructure style, and I actually like it as a place to set default props, I'm just confused about the need TS-wise).

Which components were you seeing this make a difference for? And is it being caught by TS at compile time or by a JS error at runtime? If it's just helping for components which don't have types specified like `React.FunctionComponent<IProps>` at the top, then we should also/instead add that, I would expect that to be able to catch these issues at compile time. If it's helping for components which already have that kind of type parameter, I'm confused about why that would be.

gildub:
@mturley , @eriknelson , @ibolton336 

I basically got errors when I replaced props variable by { prop1, ..., propN}.

For instance, the WizardComponent calls StorageClassForm and is passing props, meanwhile `errors`, `touched`, `handleBlur`, `handleChange`, `setFieldTouched`, are not defined in StorageClassForm. Those are not optionals, there are simply not defined on StorageClassForm.

What happens is when using `const { ... } = props`, extra fields are simply ignored.

I also found `isFetchingPVResources` is needed by VolumeForm component but is not passed from WizardComponent.

mturley:
But for e.g. StorageClassForm, you were getting these errors at run-time in the browser console, right? Those are things we shouldn't have to catch at run time, TypeScript should be able to catch them for us. We should just add a <PropsInterface> to every function component.

> For instance, the WizardComponent calls StorageClassForm and is passing props, meanwhile `errors`, `touched`, `handleBlur`, `handleChange`, `setFieldTouched`, are not defined in StorageClassForm. Those are not optionals, there are simply not defined on StorageClassForm.

> What happens is when using `const { ... } = props`, extra fields are simply ignored.

That's what is confusing me. Both syntaxes behave the same in terms of ignoring fields. Destructuring in the function argument list and destructuring in the function body have the same behavior. I can verify that in my console with a plain function:

```
const fn1 = ({ foo, bar }) => console.log({ foo, bar });
fn1({ foo: 1, bar: 2, baz: 3 }); // logs { foo: 1, bar: 2 }, no errors because of unexpected "baz"

const fn2 = props => {
  const { foo, bar } = props;
  console.log({ foo, bar });
};
fn2({ foo: 1, bar: 2, baz: 3 }); // logs { foo: 1, bar: 2 }, no errors because of unexpected "baz"
```

Can you post the text of the error you were getting?
```

**Comment by mturley** [comment]
I think maybe the best way to deal with this is to both do what you've done destructuring props in the function arguments, but also add a props interface to every component and specify it both in the `React.FunctionComponent<>` type annotation and also as a type annotation on the destructured object. This is what we're doing internally in PatternFly: https://github.com/patternfly/patternfly-react/blob/master/packages/react-core/src/components/Button/Button.tsx#L55-L74 (look at the two places `ButtonProps & InjectedOuiaProps` is specified). We can also use default parameters in that destructured object to specify default props wherever they are needed, like this example shows.

**Comment by eriknelson** [comment]
> I think maybe the best way to deal with this is to both do what you've done destructuring props in the function arguments, but also add a props interface to every component and specify it both in the React.FunctionComponent<> type annotation and also as a type annotation on the destructured object. 

+1. I see two problems here:

1) We're inconsistent with the way we unwrap a props object. I see this as a stylistic issue, as @mturley pointed out, destructuring in the function's signature or from within the body doesn't matter. You will still end up with undefined function scoped variables if it doesn't exist on the object (ignoring the TS annotations).

2) We should push prop type validations to compile-time using Typescript interfaces; after all that's kind of the point of using TS at all.

I would also maintain prop type interfaces should use an `I` prefix, and a `Props` suffix.

`INamespaceTableProps`, not `NamespaceTable` or some other combination. This is a reasonably well established convention.

I'm still having trouble seeing how this PR appreciably improves the current codebase. It feels like a side step rather than a step forwards, and I'd rather not make a change for the sake of it.

**Comment by mturley** [comment]
> I'm still having trouble seeing how this PR appreciably improves the current codebase. It feels like a side step rather than a step forwards, and I'd rather not make a change for the sake of it.

+1, but thank you Gilles for bringing attention to the underlying problem of props not being well-validated. I think if we add the type annotations and standardize the `IWidgetProps` naming convention this will be useful.

**Comment by gildub** [comment]
@mturley, you're welcome. BTW this is not happening at runtime but when changing from props to destructuring directly.

**Comment by gildub** [comment]
@eriknelson, 
* I disagree, destructuring is not just a style issue. It shows validation issues prone to bugs and it's also a maintenance concern.
Removing a field (for instance `isFetchingPVResources` from caller, breaks TS:
```
TS2741: Property 'isFetchingPVResources' is missing in type
``` 
This is the main reason I started this PR. I understand it involves styling and naming too meanwhile the core issue remains, using intermediate variable to destructure shouldn't be used.

* No worries keeping type names, I understand it's a convention. Naming is always something difficult.

**Comment by mturley** [comment]
@gildub,

This must be something about TS that I wasn't familiar with, because in regular JS these two forms of destructuring are equivalent and behave identically. Apparently TS is able to implicitly assume your destructuring represents a type. I do think it's better to be explicit, however. Specifying a type in `React.FunctionComponent<>` should have the same effect, no? We should definitely do that, the errors will be a little more clear since they will contain the interface name. Whether or not we also move the destructuring to the function declaration is a matter of style (I'm in favor, though).

Would you both be okay with doing this the way it's done in PF? Belt-and-suspenders, nice and explicit?

```ts
const MyComponent: React.FunctionComponent<IMyComponentProps> = ({
  propA,
  propB,
  propC
}: IMyComponentProps) => {
  // component render logic
};
```

For what it's worth, @eriknelson if the style concern is that we want to be able to pull extra props out and pass them along without refactoring, we can do that in this syntax by adding `...props` in that destructuring object which will declare `props`, which will contain all props not specified. The other benefit of this is we can specify default prop values with `=` right in the same object.

gildub:
@mturley, +1 - I think the PF way is rock solid.

eriknelson:
> Apparently TS is able to implicitly assume your destructuring represents a type.

Unless the function has been annotated it shouldn't be using type inference. I'm still not understanding what @gildub is seeing, maybe a specific example can help clear that up.

mturley:
@eriknelson I'm able to reproduce what @gildub was seeing, apparently even if you don't specify types, TS is smart enough to assume everything you destructure in a function signature is required unless it has a default value. It just implicitly uses `any` for the types of those properties.

Screenshot 2020-03-31 11 52 59

Screenshot 2020-03-31 11 55 09

gildub:
@ibolton336, per scrum discussion we are going to go ahead and apply this gradually using PF type style.

gildub:
@mturley, I believe if no type is specified then any is used by default.

mturley:
@gildub yep. That's better than having no validation of required arguments, but we should avoid using the `any` type whenever possible.

mturley:
In fact, in the future we might want to add the linter rule `no-implicit-any`

eriknelson:
@gildub can you please update anything this PR originally touched to use the prop type and explicitly annotate the types. I think we can merge it at that point.

mturley:
@gildub @eriknelson would you guys be ok with this PR not modifying `VolumesForm.tsx` or `VolumesTable.tsx`? I'm adding types to the props for those in https://github.com/konveyor/mig-ui/pull/776 and moved them to this new style. There may be other minor conflicts but this would prevent the significant ones.

eriknelson:
@mturley +1, makes sense.

gildub:
@mturley, no pb, leaving VolumeForm component, including it's call untouched for now.

mturley:
Thanks @gildub. Also you can see what I mentioned on standup about reusing types from other interfaces in this commit: https://github.com/konveyor/mig-ui/pull/776/commits/dec93f34a3140e8900284265ceb0232b0248710a

The `extends Pick<...>` syntax takes another interface and a `|`-separated condition for which props match, and picks those props to also be in your interface.

I'm thinking maybe we should do this at first for most of the reused props in the wizard, and then look over that to see if any groups of props that are always passed together should be moved to their own interfaces and extended like @eriknelson mentioned (or even pass those grouped props as one object prop to each component).

mturley:
Details about TS utility types here: https://www.typescriptlang.org/docs/handbook/utility-types.html

`Omit<...>` is also useful.

mturley:
@gildub can you also restore `VolumesTable` and its call for now? My PR changes those props around and adds types there too.

gildub:
@mturley, I restored `VolumesTable too`. Thanks for the links

gildub:
@eriknelson, @mturley, and @ibolton336, I've forced pushed as I had to rebase upon PR#776 and I had to re-visite my initial approach.

src/app/plan/components/Wizard/WizardContainer.tsx
```diff
@@ -13,6 +13,7 @@ export interface IFormValues {
   selectedStorage: string;
   selectedNamespaces: any[];
   persistentVolumes: any[]; // TODO replace this with selections-only version after refactor
+  pvStorageClassAssignment: any[];
```
mturley, line 16: Let's try to avoid using `any` when we can (we should eventually make better types for all of these `any` props). In this case it looks like `pvStorageClassAssignment` is an object with arbitrary string keys that map to string values.
```
pvStorageClassAssignment: {
  [key: string]: string;
}
```

mturley: Also I hadn't noticed the `[]` on `any[]`. I think it's an object, not an array

src/app/plan/components/Wizard/GeneralForm.tsx
```diff
@@ -11,15 +13,25 @@ interface IProps {
...
 }
 
-const GeneralForm: React.SFC<IProps & RouteComponentProps> = ({
-  handleChange,
-  handleBlur,
-  values,
+interface IGeneralFormProps
+  extends Pick<
+    IOtherProps,
+    | 'isEdit'
```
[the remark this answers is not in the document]
gildub: Yes absolutely.

src/app/plan/components/Wizard/GeneralForm.tsx
```diff
@@ -11,15 +13,25 @@ interface IProps {
...
+    >,
+    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
+    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
+    Pick<FormikState<IFormValues>, 'touched'> {
+      errors: any;
+    }
+const GeneralForm: React.FunctionComponent<IGeneralFormProps & RouteComponentProps> = ({
```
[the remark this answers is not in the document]
gildub: Legacy code has is using `React.SFC<IProps & RouteComponentProps>`.
Not sure why it's there either.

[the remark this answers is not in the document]
gildub: Let me remove the RouteComponentProps then.

src/app/plan/components/Wizard/StorageClassForm.tsx
```diff
@@ -6,9 +6,28 @@ import {
...
+
+interface IStorageClassFormProps
+  extends Pick<
+    IOtherProps,
+    | 'clusterList'
+    | 'currentPlan'
+    | 'isEdit'
+    | 'isFetchingPVList'
+    >,
+    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'> {}
```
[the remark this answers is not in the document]
gildub: I agree.

src/app/plan/components/Wizard/WizardContainer.tsx
gildub: This doesn't work.
Although I agree with the target to get rid of `any` it seems there is still lots of work to get there.
All the properties above are any[] too.
I believe it should be addressed separately.

mturley: I think the type errors you were getting were because the `mapStateToProps` function that initializes the values was also passing in `[]` instead of `{}`. I think I may address this in my next PR for the storage class table anyway though.

src/app/plan/components/Wizard/GeneralForm.tsx
```diff
@@ -14,21 +13,29 @@ interface IProps {
...
-    Pick<FormikActions<IFormValues>, 'setFieldTouched'>,
-    Pick<FormikProps<IFormValues>, 'setFieldValue' | 'values'>,
-    Pick<FormikState<IFormValues>, 'touched'> {
+  extends Pick<
+    IOtherProps, 'isEdit'
+    >,
+    Pick<FormikProps<IFormValues>, 
+    | 'handleBlur'
+    | 'handleChange'
+    | 'setFieldTouched'
```
mturley, line 22: This is fine for this PR, but I think Prettier lays it out differently. @ibolton336 we should figure out a good time to run Prettier on the whole repo (and merge it quick so we don't cause conflicts). @gildub I would recommend installing the Prettier plugin for VS Code and letting it format your code on save (or if you disable that, you can tell it to format a selection).

mturley:
Hmm, @eriknelson @ibolton336 do you guys know why Travis CI might have timed out here? I wonder if it's not unique to this PR, I don't see anything problematic in here.

eriknelson:
Usually I can restart builds to force it to run again, but for some reason I'm not seeing that as an option in the UI. @gildub might be able to do it, but a hacky workaround is to just close and reopen.

eriknelson:
@mturley @gildub looks like kicking travis fixed the job

gildub:
@mturley, now I'm using Prettier plugin and I've pushed reformatted files.

eriknelson:
@ibolton336 can you re-review?

gildub:
@ibolton336, I've adjusted after rebase.

mturley:
@ibolton336 we should merge this sooner rather than later since it's touching a handful of files in places that may cause conflicts.

PR #784, opened 2020-03-30
Fixes regression with delete plan modal

ibolton336:
Adds back in call to delete plan

### DIFF
--- a/src/app/home/components/DataList/Plans/PlanActions.tsx
+++ b/src/app/home/components/DataList/Plans/PlanActions.tsx
@@ -136,13 +136,17 @@ const PlanActions = ({ plan, history }) => {
           plan={plan}
           isOpen={isMigrateModalOpen}
           onHandleClose={toggleMigrateModalOpen}
-        /> 
+        />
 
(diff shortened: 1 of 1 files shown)


src/app/home/components/DataList/Plans/PlanActions.tsx
```diff
@@ -136,13 +136,17 @@ const PlanActions = ({ plan, history }) => {
...
           onHandleClose={toggleMigrateModalOpen}
-        /> 
+        />
 
         <ConfirmModal
           title="Confirmation"
           message={`Do you really want to delete migration plan "${plan.MigPlan.metadata.name}"?`}
           isOpen={isDeleteModalOpen}
-          onHandleClose={toggleDeleteModalOpen}
+          onHandleClose={() => {
```
mturley: @ibolton336 this will delete the plan even if the user clicks Cancel. The onHandleClose function takes a boolean `isConfirmed`, we just need to use that. I put up a PR too, sorry for the conflict: https://github.com/konveyor/mig-ui/pull/785

PR #792, opened 2020-04-01
Prevent plan from setting spec.closed on migration end

### LINKED ISSUE #753: Revisit closing migplan after migration succeeds
After some discussion on the triage meeting, it sounded like there were some disagreements about how the UI closes the migplan after a migration has completed. This is done to tell the controller to clean up the associated resources rather than simply deleting the migplan which won't properly clean everything up.

@jortel had brought up that there may be a better way to solve this rather than distinguishing "closing" the plan vs deleting it.

Danil-Grigorev:
A final migration no longer closes plan on completion.

Close #753

### DIFF
--- a/src/app/home/components/DataList/Plans/PlanStatus.tsx
+++ b/src/app/home/components/DataList/Plans/PlanStatus.tsx
@@ -31,10 +31,10 @@ const PlanStatus: React.FunctionComponent<IProps> = ({ plan }) => {
       status.text = conflictErrorMsg || 'Conflicting plan';
     } else if (hasRunningMigrations) {
       status.text = `${latestType} Running`;
-    } else if (hasSucceededStage) {
-      status.text = `Stage Succeeded`;
     } else if (hasSucceededMigration) {
       status.text = `Migration Succeeded`;
+    } else if (hasSucceededStage) {
+      status.text = `Stage Succeeded`;
     } else if (hasNotReadyCondition || !hasReadyCondition) {
       status.text = 'Not Ready';
     } else if (hasReadyCondition) {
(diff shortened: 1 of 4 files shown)


eriknelson:
There has been no discussion and no consensus on whether or not this is something we want to do. We may want to merge this, but that conversation needs to happen first.

eriknelson:
Discussed on controller scrum, conclusion was that we'd like to accept this PR and remove the automatic close of plans at the conclusion of final migration. We'd also like to improve the UX by adding a checkbox to the migration pop-up/wizard when you're about to run a final migration that prompts the user whether or not they would like to automatically close the PR upon completion, and why they would want to do that vs. not. @vconzola what are your thoughts on that?

eriknelson:
@sseago I think the intent with the user opt-in is that it's also an opportunity to describe all of these cleanup operators that happen, so the user can make an educated decision themselves whether or not they think that's something desirable, and we default to the less destructive option, which is not performing any of the cleanup, since the user can always close at their leisure.

sseago:
I guess what concerns me is that it's possible for there to be a situation where the moment that a user is first notified of a potential problem is the same moment that we remove the user's ability to learn more about what happened. As long as we're clear "selecting this might prevent you from investigating issues that you won't be informed about until it's too late", then I guess that's OK.

vconzola:
@eriknelson What's the use case for **_not_** closing a mig plan after final migration? What state would such a plan be in and what actions would a user be able to take on it? It's easy enough to add the checkbox you describe to the migration popup where we ask about quiescing workloads on the source cluster. I assume we'd also need to add a Close action to the plan kebab menu for plans that have been migrated successfully but not closed.

eriknelson:
Visually this LGTM, I want to do an integration test before merging but I think we can get this merged today. I have a call I need to get on in 10 minutes but I will try to do that before EOD.

PR #801, opened 2020-04-07
Adds Prettier plugin to ESLint

gildub:
* Replaces legacy `.prettierrc.json` to JS file `.prettierrc.js` (same as `.eslintrc.js`, having a JS file is more flex allowing comments, etc)
* Applies `yarn eslint . --ext .js,.jsx,.ts,.tsx  --fix`
* Adds React plugin to ESLint but not activated yet

For more details please check out https://prettier.io/docs/en/integrating-with-linters.html

Also there is a comprehensive description to help us get started: https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project

Closes https://github.com/konveyor/mig-ui/issues/775


ibolton336:
Needs a rebase

gildub:
@ibolton336, rebase done.

mturley:
Now that https://github.com/konveyor/mig-ui/pull/802 is merged this one is next, right @eriknelson @ibolton336? @gildub when you get a chance to rebase/update this, I think we're holding other PRs so it can be merged

.eslintrc.js
```diff
@@ -1,29 +1,37 @@
 module.exports = {
     root: true,
     parser: '@typescript-eslint/parser',
-    plugins: [
-        '@typescript-eslint',
-    ],
+    plugins: ['@typescript-eslint'],
```
mturley, line 4: I think we also need to add `'prettier'` to the plugins array here, and `'prettier/prettier': 'error'` to the rules object below.

From https://github.com/prettier/eslint-plugin-prettier:
```
{
  "plugins": ["prettier"],
  "rules": {
    "prettier/prettier": "error"
  }
}
```

mturley: I think we may also want to add https://github.com/yannickcr/eslint-plugin-react. If we do, we'll have to add it and the prettier plugin line for compatibility with it.
In `plugins`: `'react'`
In `extends`: `plugin:react/recommended` and `prettier/react`

gildub: @mturley, actually https://github.com/prettier/eslint-plugin-prettier#recommended-configuration says that adding `plugin:prettier/recommended` automatically sets `prettier/prettier:error` in rules.

Also the react plugin is going to wait once we have prettier in place because that's will require to fix lots of files.

mturley: Ah got it, I didn't see that! Perfect, thanks.

Sounds good on the React plugin. We should follow up on that, there are some really useful rules in there.

PR #802, opened 2020-04-08
Feature: Add/remove/update hooks from the UI

### LINKED ISSUE #668: Pre/post ansible hooks
https://issues.redhat.com/browse/MIG-50

~Also blocked on non-admin engineering design.~

ibolton336:
Closes #668


ibolton336:
> @ibolton336 Would you please add either a series of screenshots or a video showing UI flow to the PR. Thanks.

add_hook

vconzola:
@ibolton336 This looks good. There are a number of little spacing and capitalization tweaks that are needed to clean it up. But behaviorally the only thing I noticed is the way the hooks are added/updated from the modal. I added a few pages to the [mockup document](https://docs.google.com/presentation/d/1WgrJ1B4J1nmgTaHW7vJq-xbgjDS8kPIQq4IBpABzvKg/edit?usp=sharing) that detail the required changes. We can talk about them in scrum tomorrow.

ibolton336:
Screen Shot 2020-04-15 at 11 26 50 AM

eriknelson:
Failing the linter on my machine due to indents looks like

ibolton336:
Rebased. 
Oustanding issues:

-Need to update the hooks form to follow the same pattern used on add/edit cluster/storage form. This includes adding a status poll & a watcher for the hook object & associated plan that will identify any error conditions related to the hook add/update request.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+                setFieldValue('migrationStep', value);
+                setFieldTouched('migrationStep');
+              }}
+              options={phaseOptions}
+              value={values.migrationStep}
+              placeholderText="Select phase..."
+            />
+          </FormGroup>
+        </GridItem>
+        <Grid span={6}>
```
mturley, line 414: We can leave it, but I wonder if we're overusing `Grid` in general. Two buttons directly in a div with some spacer classes on them would probably do fine here.

src/app/plan/components/Wizard/HooksFormContainer.tsx
```diff
@@ -0,0 +1,87 @@
+import { withFormik, FormikProps } from 'formik';
+import utils from '../../../common/duck/utils';
+import HooksFormComponent from './HooksFormComponent';
+
+const AddEditHooksFormContainer: any = withFormik({
```
mturley, line 5: The `any` type annotation is probably not doing us any favors here. At the very least it should be `React.FunctionComponent<any>`, but ideally we'd have a full props interface here. Looks like we would just need to import `IOtherProps` from `HooksFormComponent`.

Also, we should pass generic types into `withFormik` like we do in `WizardContainer`, e.g. `withFormik<IOtherProps, IFormValues>(...)`. That'll typecheck the values we return in here.

src/app/plan/components/Wizard/HooksFormContainer.tsx
```diff
@@ -0,0 +1,87 @@
...
+      customContainerImage: 'quay.io/konveyor/hook-runner:latest',
+      ansibleRuntimeImage: 'quay.io/konveyor/hook-runner:latest',
+      clusterType: 'source',
+      srcServiceAccountName: '',
+      destServiceAccountName: '',
+      srcServiceAccountNamespace: '',
+      destServiceAccountNamespace: '',
+      migrationStep: '',
+      ansibleFile: ''
+    };
```
mturley, line 19: We should also maybe add an annotation here, `const values = { ... }: IFormValues;` so that TS can help us with missing fields / bad types here. We should probably do the same in the WizardContainer tbh. There were some object-type fields that were initialized with empty arrays there that I found in my earlier work.

src/app/plan/components/Wizard/HooksStep.module.scss
```diff
@@ -0,0 +1,8 @@
+@import "src/colors";
+
+.formContainerStyles {
```
mturley, line 3: Do we really need these? the margin and padding can probably be done with the PF spacer utility classes, and I'm not sure we need the light gray border around the form. @vconzola how do you feel about that grey box? I'm assuming it was added because there was a box around the form in your mocks, but was that just a structural thing or did you intend for it to have a border? I don't think any of our other forms have this kind of border.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { SearchIcon } from '@patternfly/react-icons';
+import {
+  AddEditMode,
+} from '../../../common/add_edit_state';
+import HooksFormContainer from './HooksFormContainer';
+
+const classNames = require('classnames');
+const styles = require('./HooksStep.module')
+
+const HooksStep = (props) => {
```
mturley, line 39: We should define an interface for the props here, give this type annotations and destructure props in the function sig.

```ts
const HooksStep: React.FunctionComponent<IHooksStepProps> = ({
  ...
}: IHooksStepProps) => {
  ...
};
```

mturley: Seems like a lot of the props here are the same as the hooks form components, so maybe those types could either be defined down in the form and extended here, or defined here and `Pick<...>`'d down in the form.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+    fetchHooksRequest,
+    hookAddEditStatus,
+    currentPlan,
+    removeHookRequest,
+    watchHookAddEditStatus,
+    isAddHooksOpen,
+    setIsAddHooksOpen
+  } = props;
+
+  const [initialHookValues, setInitialHookValues] = useState({});
```
mturley, line 54: Take it or leave it (we don't really do this in many places, but we probably should): we can add a generic type annotation to `useState` for these complex state values. e.g. `useState<IMigHook>({});`.

Simple `useState` calls with a primitive type in their default value (string, number, boolean) don't need these because TS can infer the type from the initial value and use that to check your usage, but when the default value is an empty object or array we should probably be passing generics like this.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+      {
+        title: 'Delete',
+        onClick: (event, rowId, rowData, extra) => {
+          removeHookRequest(rowData.name.title, rowData["migration-step"].title)
+        }
+      }
+    ]
+
+  }
+  if (rows.length === 0) {
```
mturley, line 117: This is fine, but it might be cleaner to just render the empty state instead of the table altogether when there are no rows, instead of creating a placeholder row. I do that in the other tables (https://github.com/konveyor/mig-ui/blob/master/src/app/plan/components/Wizard/VolumesTable.tsx#L188) and I created a reusable `TableEmptyState` component for it.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+            initialHookValues={initialHookValues}
+            setInitialHookValues={setInitialHookValues}
+            setIsAddHooksOpen={setIsAddHooksOpen}
+            isAddHooksOpen={isAddHooksOpen}
+            isEditHook={false}
+            {...props} />
+        </GridItem>
+      }
+      {(!isAddHooksOpen) &&
+        <React.Fragment>
```
mturley, line 168: I don't think you need a fragment here, you can just have the `isFetchingHookList ? <Bullseye> : <GridItem>` expression directly in the place of the fragment. You only need a fragment if you need more than one top-level node in your JSX expression.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+                <Title headingLevel="h2" size="xl">
+                  Finding hooks for this migration plan...
+                </Title>
+              </EmptyState>
+            </Bullseye>
+            :
+            <GridItem className={hooksFormContainerStyles}>
+              <Grid gutter="md">
+                <GridItem span={4}>
+                  <Button
```
mturley, line 184: In the mocks it looks like there is a search filter to the left of the "Add hook" button. @vconzola did we want filters, sorting, and pagination in this table? @ibolton336 we can do that in a followup PR, but let me know if you want me to explain the `useFilterState` / `useSortState` / `usePaginationState` stuff.

src/app/plan/components/Wizard/HooksStep.tsx
```diff
@@ -0,0 +1,220 @@
...
+    }]
+  }
+
+  const hooksFormContainerStyles = classNames(spacing.mtSm, styles.formContainerStyles)
+  return (
+    <Grid>
+      <GridItem className={spacing.mtSm}>
+        <TextContent>
+          <Text component={TextVariants.h2}>
+            Hooks
```
mturley, line 144: We can actually remove this Hooks text header and use `<WizardStepContainer>` in the `WizardComponent` instead.

src/app/plan/components/Wizard/WizardComponent.tsx
```diff
@@ -173,7 +187,33 @@ const WizardComponent = (props: IOtherProps & FormikProps<IFormValues>) => {
             </WizardStepContainer>
           ),
           canJumpTo: stepIdReached >= stepId.StorageClass,
-          nextButtonText: 'Finish',
+        },
+        {
+          id: stepId.Hooks,
+          name: 'Hooks',
+          component: (
+            <HooksStep
```
mturley, line 195: If you wrap this in `<WizardStepContainer title="Hooks">` you don't need the title text inside HooksStep.

src/app/plan/components/Wizard/WizardComponent.tsx
```diff
@@ -207,10 +247,15 @@ const WizardComponent = (props: IOtherProps & FormikProps<IFormValues>) => {
       errors,
       touched,
       currentPlanStatus,
+      migHookList,
```
mturley, line 250: This big useEffect is wild, I might take a closer look at it later 😆

src/app/plan/components/Wizard/WizardComponent.tsx
```diff
@@ -233,6 +278,7 @@ const WizardComponent = (props: IOtherProps & FormikProps<IFormValues>) => {
           targetCluster: props.values.targetCluster,
           selectedStorage: props.values.selectedStorage,
           namespaces: props.values.selectedNamespaces,
+          hookDefinitions: props.values.hookDefinitions
```
mturley, line 281: We should probably add an interface to define the types of this argument object on `addPlanRequest`. Maybe it's just `IFormValues`?

src/app/plan/components/Wizard/WizardComponent.tsx
```diff
@@ -241,6 +287,10 @@ const WizardComponent = (props: IOtherProps & FormikProps<IFormValues>) => {
       //update plan & start status polling on results page
       planUpdateRequest(props.values, false);
     }
+    if (prev.prevId === stepId.Hooks && curr.id === stepId.StorageClass) {
```
mturley, line 290: Another use case for https://github.com/patternfly/patternfly-react/issues/4104, we wouldn't need this `isAddHooksOpen` state if we just had `currentStep` lifted out of `Wizard`.

src/app/plan/components/Wizard/WizardComponent.tsx
mturley: In that case, I wonder if we can just pass `props.values` instead of repeating all its properties here

src/app/plan/components/Wizard/WizardContainer.tsx
```diff
@@ -32,6 +40,7 @@ export interface IFormValues {
   pvCopyMethodAssignment: {
     [pvName: string]: PvCopyMethod;
   };
+  hookDefinitions: any[];
```
mturley, line 43: This can just be `IMigHook` right?

src/app/plan/components/Wizard/WizardContainer.tsx
```diff
@@ -66,13 +75,23 @@ export interface IOtherProps {
...
+  fetchHooksRequest: (currentPlanHooks) => void;
   addPlanRequest: (migPlan) => void;
   sourceClusterNamespaces: ISourceClusterNamespace[];
   pvResourceList: IPersistentVolumeResource[];
   onHandleWizardModalClose: () => void;
   editPlanObj?: any;
   isEdit: boolean;
   updateCurrentPlanStatus: any;
+  addHookRequest: (migHook) => void;
+  updateHookRequest: (migHook) => void;
```
mturley, line 87: Ideally each of these callback arguments would have type annotations too, they are implicitly `any` here

src/app/plan/components/Wizard/WizardContainer.tsx
```diff
@@ -97,6 +129,7 @@ const WizardContainer = withFormik<IOtherProps, IFormValues>({
       // TODO need to look into this closer, but it was resetting form values after pv discovery is run & messing with the UI state
       // See https://github.com/konveyor/mig-ui/issues/797
       // values.persistentVolumes = editPlanObj.spec.persistentVolumes || [];
+      values.hookDefinitions = editPlanObj.spec.hookDefinitions || [];
```
mturley, line 132: Ah, we should probably add `hookDefinitions` to the `IPlan` interface here: https://github.com/konveyor/mig-ui/blob/master/src/app/plan/components/Wizard/types.ts#L35-L49 and then reuse the new HookDefinition interface wherever we use these values

src/app/plan/duck/reducers.ts
```diff
@@ -341,6 +345,117 @@ export const planCloseAndDeleteFailure = (state = INITIAL_STATE, action) => {
   return unlockPlan(state, action.planName);
 };
 
+/* 
+Hook Reducers
+*/
+
+export const removeHookRequest =
+  (state = INITIAL_STATE, action: ReturnType<typeof PlanActions.removeHookRequest>) => {
```
mturley, line 353: I think these `ReturnType<...>` annotations are only useful if the types of the variables being returned in the referenced function are well-defined. It might be better to be explicit about all these types wherever we use `ReturnType`, but that's a much bigger change. We should at least try to put an explicit return type annotation on the action functions.

src/app/plan/duck/reducers.ts
```diff
@@ -34,7 +35,10 @@ export const INITIAL_STATE = {
     state: CurrentPlanState.Pending,
     errorMessage: ''
   } as ICurrentPlanStatus,
-  lockedPlanList: []
+  lockedPlanList: [],
+  isFetchingHookList: false,
+  migHookList: [],
+  hookAddEditStatus: defaultAddEditStatus()
 };
```
mturley, line 42: In the future we should probably make type interfaces for the initial state in each of these reducers

src/app/plan/duck/sagas.ts
```diff
@@ -693,6 +719,274 @@ function* migrationPoll(action) {
...
+          const destServiceAccountName = hookRef.spec.targetCluster === 'destination' ? currentPlanHookRef.serviceAccount : null;
+          const destServiceAccountNamespace = hookRef.spec.targetCluster === 'destination' ? currentPlanHookRef.executionNamespace : null;
+          const clusterTypeText = hookRef.spec.targetCluster === 'destination' ? 'Target cluster' : 'Source cluster';
+
+          let ansibleFile;
+          if (!hookRef.spec.custom) {
+            ansibleFile = atob(hookRef.spec.playbook);
+          }
+
+          const uiHookObject = {
```
mturley, line 755: I think this would be another usage of that HookDefinition type interface

src/app/plan/duck/sagas.ts
```diff
@@ -693,6 +719,274 @@ function* migrationPoll(action) {
   }
 }
 
+/******************************************************************** */
+//Hooks sagas
+/******************************************************************** */
+
+function* fetchHooksSaga(action) {
```
mturley, line 726: This looks good to me, but I need to learn more about sagas and come back to look at all of these.

src/client/resources/conversions.ts
```diff
@@ -508,6 +508,164 @@ export function createMigMigration(
   };
 }
 
+export function updateMigHook(
+  currentHook: any,
+  migHook: any,
+  currentPlanHookRef: any,
+  namespace: string,
+  currentPlan: any
```
mturley, line 516: I think `currentPlan` could reuse the `IPlan` interface from the wizard `types.ts`, although maybe we should be defining some of those types in a more common place. These other hook `any` types should be replaced too if possible

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+    errors,
+    setFieldValue,
+    setFieldTouched,
+    values,
+    handleChange,
+    handleBlur,
+    cancelAddEditWatch,
+    resetAddEditState,
+    currentPlan
+  } = props;
```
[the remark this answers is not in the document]
ibolton336: +1 done

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+  customContainerImage: string;
+  srcServiceAccountName: string;
+  srcServiceAccountNamespace: string;
+  destServiceAccountName: string;
+  destServiceAccountNamespace: string;
+  clusterType: string;
+  migrationStep: string;
+}
+interface IOtherProps {
+  onAddEditSubmit: any;
```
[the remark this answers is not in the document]
ibolton336: +1 done. Need to take a second pass and really remove all any types. Maybe even enforce it in the eslint rules.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+  ansibleRuntimeImage: string;
+  customContainerImage: string;
+  srcServiceAccountName: string;
+  srcServiceAccountNamespace: string;
+  destServiceAccountName: string;
+  destServiceAccountNamespace: string;
+  clusterType: string;
+  migrationStep: string;
+}
+interface IOtherProps {
```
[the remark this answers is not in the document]
ibolton336: +1 done

src/app/plan/components/Wizard/HooksForm.module.scss
```diff
@@ -0,0 +1,3 @@
+.tooltipStyles {
+    margin-left: .5em !important;
```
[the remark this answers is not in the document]
ibolton336: +1 done

package.json
```diff
@@ -62,7 +62,7 @@
   },
   "dependencies": {
     "@patternfly/react-core": "^3.146.0",
-    "@patternfly/react-table": "^2.28.10",
+    "@patternfly/react-table": "^2.25.6",
```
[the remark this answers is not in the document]
ibolton336: +1 done

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+    handleBlur,
+    cancelAddEditWatch,
+    resetAddEditState,
+    currentPlan
+  } = props;
+
+  const formikHandleChange = (_val, e) => handleChange(e);
+  const formikSetFieldTouched = key => () => setFieldTouched(key, true, true);
+
+  const hookNameKey = 'hookName';
```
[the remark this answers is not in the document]
ibolton336: I think it makes sense to leave these here. Otherwise we are risking typos when retyping the string for fieldId, name, and anywhere it is passed in as a param to the formik value change wrapper functions. I vote to keep them for now. This is definitely up for discussion, though. There might be a way to clean all of this up including the formik wrapper functions. Not sure if those are still needed.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+    'PostRestore'
+  ]);
+
+  useEffect(() => {
+    if (currentPlan.spec.hooks) {
+      const existingPhases = currentPlan.spec.hooks.map((hook) => {
+        return hook.phase
+      })
+      let filteredPhases = []
+      filteredPhases = phaseOptions.filter(phase => !existingPhases.includes(phase));
```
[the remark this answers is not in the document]
ibolton336: +1

src/app/plan/components/Wizard/HooksStep.tsx
eriknelson: Think this is a good candidate for an issue and a follow up PR.

src/app/plan/components/Wizard/HooksFormComponent.tsx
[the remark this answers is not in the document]
eriknelson: > I think it makes sense to leave these here. Otherwise we are risking typos when retyping the string for fieldId, name, and anywhere it is passed in as a param to the formik value change wrapper functions.

+1, this is why I defined these as consts. Always prefer const defs for hooks rather than string literals that can be botched. Unless I'm misunderstanding some way that typescript will prevent this.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+  const destServiceAccountNamespaceKey = 'destServiceAccountNamespace';
+  const migrationStepKey = 'migrationStep'
+  const [phaseOptions, setPhaseOptions] = useState([
+    'PreBackup',
+    'PostBackup',
+    'PreRestore',
+    'PostRestore'
+  ]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
ibolton336: +1 nice! This works great.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+                      isRequired
+                      fieldId={ansibleRuntimeImageKey}
+                      isValid={!(touched.ansibleRuntimeImage
+                        && errors.ansibleRuntimeImage)}
+                    >
+                      {/*
+          // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
+                      <TextInput
+                        onChange={formikHandleChange}
+                        onInput={formikSetFieldTouched(ansibleRuntimeImageKey)}
```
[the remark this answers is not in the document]
ibolton336: This one might be a good one to address on a global app level. I agree I think these aren't necessary.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+                  name={hookImageTypeKey}
+                  onChange={formikHandleChange}
+                  label="Ansible playbook"
+                  id="ansible-playbook-radio"
+                  value="ansible"
+                />
+              </GridItem>
+              {values.hookImageType === 'ansible' &&
+                <React.Fragment>
+                  <GridItem className={hookImageStyles}>
```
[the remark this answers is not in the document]
ibolton336: const hookImageStyles = classNames(spacing.mtSm, spacing.mlMd);

Using this for styling here.

src/app/plan/components/Wizard/HooksFormComponent.tsx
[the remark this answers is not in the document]
mturley: I'm fine with that, especially because I'm not 100% sure the formik typings support this. But in general if a function argument has a type that is an enum or a specific set of values (e.g. `foo: 'a' | 'b' | 'c'`) TypeScript will catch a string literal that doesn't match, and IntelliSense will autocomplete string literals from the valid values. I would hope formik uses the `IFormValues` keys to populate that argument's type. Let's leave this though, nothing wrong with a little more confidence.

src/app/plan/components/Wizard/HooksFormComponent.tsx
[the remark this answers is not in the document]
mturley: Maybe! We should at least use the `no-implicit-any` rule, so we are forced to put `any` in places where we don't have any type annotation, and they will stick out more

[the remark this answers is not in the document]
eriknelson: I like the `no-implicit-any` rule, maybe this is another candidate for an issue? I'm assuming there will be quite a bit of churn to be able to pass that.

src/app/plan/components/Wizard/HooksFormComponent.tsx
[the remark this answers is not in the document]
mturley: Ah ok, I think I mixed things up then.

src/app/plan/components/Wizard/HooksFormComponent.tsx
ibolton336: +1 lets open a seperate issue to fix this app wide. Agree that we are using it too much & probably incorrectly.

vconzola:
@mturley You asked above about sort/filter on the table. I discussed earlier with @ibolton336 - the plan is to eventually move this table out of the wizard when Hooks become a first class resource. At that time, we'll want sorting/filtering because there could be a large number of hooks. We don't need it now because a single plan can have at most 4 hooks. But if it's easier to add it now rather than later, go for it.

src/app/plan/components/Wizard/HooksStep.module.scss
ibolton336: I will remove the custom styles here.

src/app/plan/components/Wizard/HooksStep.tsx
ibolton336: +1 done

src/app/plan/duck/reducers.ts
ibolton336: +1 candidate for a global issue

src/app/plan/components/Wizard/WizardContainer.tsx
ibolton336: Removed hookDefinitions as it was unused.

src/app/plan/components/Wizard/WizardContainer.tsx
ibolton336: removed unused hookDefinitions

src/app/plan/components/Wizard/WizardComponent.tsx
ibolton336: removed hookDefinitions as it was unused. We should revisit ts support for addPlan in a later issue.

src/app/plan/components/Wizard/WizardComponent.tsx
ibolton336: Yeah It could use another look. Initially our entire wizard was rapidly rerendering due to background polling. Now that the background polling is under control, this is another candidate for refactor that we never got around to.

src/app/plan/components/Wizard/HooksStep.tsx
ibolton336: Sure that would be great to have you walk me through that. Going to leave the filter out for this pass.

src/app/plan/components/Wizard/HooksStep.tsx
ibolton336: Add this one to the list for future TS support

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+            fieldId={migrationStepKey}
+            className={styles.styledFormGroupTitle}
+            helperTextInvalid={touched.migrationStep && errors.migrationStep}
+            isValid={!(touched.migrationStep && errors.migrationStep)}
+          >
+            <SimpleSelect
+              id="migrationStep"
+              onChange={(value) => {
+                setFieldValue('migrationStep', value);
+                setFieldTouched('migrationStep');
```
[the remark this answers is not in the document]
ibolton336: +1 need to refactor this to remove the touched calls. Not sure why we started doing that.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -0,0 +1,442 @@
...
+          <FormGroup
+            label="Hook name"
+            isRequired
+            fieldId={hookNameKey}
+            helperTextInvalid={touched.hookName && errors.hookName}
+            className={styles.styledFormGroupTitle}
+            isValid={!(touched.hookName && errors.hookName)}
+          >
+            {/*
+          // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
```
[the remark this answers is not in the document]
ibolton336: +1

src/app/plan/duck/reducers.ts
ibolton336: +1 this is used across the app. TS support would be nice.

src/app/plan/duck/sagas.ts
ibolton336: This one is tricky because I had to combine some values from MigPlan & MigHook to satisfy the ui table requirements. I am in favor of coming back and hitting this one as part of the ts support second pass.

src/app/plan/components/Wizard/HooksStep.tsx
ibolton336: Needed it in this case for the nested conditional. There is probably another way to refactor this to make it cleaner.

src/app/plan/duck/sagas.ts
```diff
@@ -693,6 +719,274 @@ function* migrationPoll(action) {
...
+    return hook.reference.name === migHook.hookName;
+  })
+  const migHookObj = updateMigHook(
+    currentHook,
+    migHook,
+    currentPlanHookRef,
+    migMeta.namespace,
+    currentPlan
+  );
+  //need to patch entire hooks array
```
eriknelson, line 940: +1 this gets into patch strategy, there are different ways you can handle the array depending on it. I'm forgetting the lower level details, but there are a few ways you can tell k8s to handle it. I set the patch requester hardcoded to use `merge-patch+json`: https://github.com/konveyor/mig-ui/blob/master/src/client/client.ts#L43

src/client/resources/discovery.ts
```diff
@@ -70,5 +70,3 @@ export class PlanPodReportDiscovery extends DiscoveryResource {
     return report.data;
   }
 }
-
```
eriknelson, line 73: File doesn't need to be changed.

src/client/resources/conversions.ts
```diff
@@ -508,6 +508,164 @@ export function createMigMigration(
...
 
+export function updateMigHook(
+  currentHook: any,
+  migHook: any,
+  currentPlanHookRef: any,
+  namespace: string,
+  currentPlan: IPlan
+) {
+  const getImage = (imageType) => {
+    if (imageType === 'ansible') {
```
eriknelson, line 519: Think this "ImageType" value is another good one to use an enum for.

src/client/resources/conversions.ts
```diff
@@ -508,6 +508,164 @@ export function createMigMigration(
...
+    custom: isCustom
+  };
+
+  const currentExecutionNamespace = currentPlanHookRef.executionNamespace;
+  const currentServiceAccount = currentPlanHookRef.serviceAccount;
+  let executionNamespaceUpdated;
+  let executionNamespaceValue;
+  let serviceAccountUpdated;
+  let serviceAccountValue;
+  if (migHook.clusterType === 'source') {
```
eriknelson, line 568: Can use the enum here.

src/app/plan/components/Wizard/HooksFormComponent.tsx
```diff
@@ -46,6 +46,27 @@ interface IHooksFormOtherProps {
...
+const customImageKey = 'customContainerImage';
+const ansibleRuntimeImageKey = 'ansibleRuntimeImage';
+const clusterTypeKey = 'clusterType';
+const srcServiceAccountNameKey = 'srcServiceAccountName';
+const destServiceAccountNameKey = 'destServiceAccountName';
+const srcServiceAccountNamespaceKey = 'srcServiceAccountNamespace';
+const destServiceAccountNamespaceKey = 'destServiceAccountNamespace';
+const migrationStepKey = 'migrationStep';
+
+export const HooksClusterType = {
```
eriknelson, line 61: +1, I remember actual typescript enums being a PITA for some reason.

src/app/plan/components/Wizard/HooksFormContainer.tsx
```diff
@@ -18,13 +18,15 @@ interface IHooksFormContainerOtherProps {
...
+const defaultHookRunnerImage = 'quay.io/konveyor/hook-runner:latest'
+
 const AddEditHooksFormContainer = withFormik<IHooksFormContainerOtherProps, IHooksFormContainerValues>({
   mapPropsToValues: ({ initialHookValues }) => {
     const values: IHooksFormContainerValues = {
       hookName: '',
       hookImageType: 'ansible',
-      customContainerImage: 'quay.io/konveyor/hook-runner:latest',
-      ansibleRuntimeImage: 'quay.io/konveyor/hook-runner:latest',
+      customContainerImage: defaultHookRunnerImage,
```
eriknelson, line 28: +1

src/app/plan/components/Wizard/HooksStep.tsx
mturley: Interesting, I would have thought nesting it in parentheses would be enough. We can revisit it later maybe. Or not 😆

src/app/plan/components/Wizard/WizardComponent.tsx
mturley: Gotcha. If we still need to guard against rerendering here, we could use `React.memo` to make it only render if certain props change. (that's the new functional equivalent to the old React class `shouldComponentUpdate` lifecycle method). https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-shouldcomponentupdate

src/app/plan/components/Wizard/HooksFormComponent.tsx
mturley: Huh. Looks like an enum should have worked fine here, I'm curious what that PITA was.

eriknelson: I think it probably would have worked better here. In my specific case, I had issues with it in the kube client where one of my goals was I wanted it all statically enforced, so for each resource I wanted them specified via enum. I might have wanted them as a static class property so you'd do something like `MigResource.MigResourceKind.MigPlan`, but I hit some weird incompatibility with it. It turned into this: https://github.com/konveyor/mig-ui/blob/master/src/client/resources/mig.ts#L19, which I wasn't entirely happy with but did the job. Now the usage is:

`const planRequest = client.create(new MigResource(MigResourceKind.MigPlan, migMeta.namespace), newMigPlan);

Example: https://github.com/konveyor/mig-ui/blob/master/src/client/resources/mig.ts#L19`

PR #807, opened 2020-04-15
Refactor results screen to show multiple conditions & GVK popover

### LINKED ISSUE #684: Need support for GVK diff warning during validation phase of MigPlan creation
Work is currently being done on providing GVK diff information on a MigPlan status condition https://github.com/fusor/mig-controller/pull/413 .

The status condition with GVK diff info between the srcCluster and destCluster will be populated after the following info has been provided / changed on the MigPlan:
 - srcCluster
 - destCluster
 - namespace list

Some initial sample output on the MigPlan would look like this:
```
  observedGeneration: 16674
  unsupportedNamespaces:
  - name: phronetic
    unsupportedResources:
    - batch/v2alpha1, Resource=cronjobs
    - batch/v2alpha1, Resource=scheduledjobs
```

^  The above indicates that these resources `cronjobs v2alpha1`, `scheduledjobs v2alpha1`, which are present in selected `phronetic` namespace on the srcCluster, have GVKs that do not exist on the destination cluster and will likely fail during migration. 

```
  - category: Warn
    lastTransitionTime: 2020-02-17T12:55:52Z
    message: 'Some namespaces are unsupported. See: `unsupportedNamespaces` for
      details.'
    status: "True"
    type: NotSupported
```

I think it makes sense for us to put a warning about these GVK diffs in the validation message section of the MigPlan wizard. We will be able to generate information on whether there are issues right after the user selects the list of namespaces to be migrated.

ibolton336:
Closes #684

### DIFF
--- a/src/app/plan/components/Wizard/ConditionItem.tsx
+++ b/src/app/plan/components/Wizard/ConditionItem.tsx
@@ -23,7 +23,17 @@ import {
   WarningTriangleIcon,
 } from '@patternfly/react-icons';
 
-const ConditionItem = ({ condition, conditionIndex, incompatibleNamespaces }) => {
+interface IConditionItemProps {
+  condition: any;
+  conditionIndex: any;
+  incompatibleNamespaces: any;
+}
+
+const ConditionItem: React.FunctionComponent<IConditionItemProps> = ({
+  condition,
+  conditionIndex,
+  incompatibleNamespaces,
+}: IConditionItemProps) => {
   return (
     <DataListItem key={conditionIndex} aria-labelledby="cluster-item">
       <DataListItemRow>
--- a/src/app/plan/components/Wizard/ConditionItem.tsx
+++ b/src/app/plan/components/Wizard/ConditionItem.tsx
@@ -81,17 +81,13 @@ const ConditionItem = ({ condition, conditionIndex, incompatibleNamespaces }) =>
                       }
                       bodyContent={
                         <>
-                          <List>
-                            <ListItem>
-                              <TextContent className={spacing.mbMd}>
-                                <Text component={TextVariants.p}>
-                                  Namespaces with GVKs that are not supported on the target cluster
-                                  will not be migrated with the rest of the migration plan. They
-                                  will need to be migrated manually, post migration.
-                                </Text>
-                              </TextContent>
-                            </ListItem>
-                          </List>
+                          <TextContent className={spacing.mbMd}>
+                            <Text component={TextVariants.p}>
+                              Namespaces with GVKs that are not supported on the target cluster will
+                              not be migrated with the rest of the migration plan. They will need to
+                              be migrated manually, post migration.
+                            </Text>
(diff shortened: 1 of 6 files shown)


ibolton336:
gvk_gif-results

ibolton336:
@vconzola Could you take a look at this when you have some time? Thanks!

ibolton336:
> @ibolton336 A couple things...
> 
> 1. Is the gvk warning that opens the popover redundant with the second warning in the list that refers users to "see 'incompatibeNamespaces' for details"? If so, we don't need both. I actually like the wording of the first warning better. Can you just replace "see 'incompatibeNamespaces' for details" with the link to the popover in that warning? If not, it's unclear what 'incompatibeNamespaces' means, so a clearer call to action is needed.
> 2. The mockups show a button link to "Return to Resources step" when the gvk warning is present. Are you able to do that? If not, the call to action in the popover needs to change because there's no way to go back to that step in the wizard from the validation screen.

+1 I will get these fixed

ibolton336:
After working with Derek, all of the controller side issues are resolved & everything is working as designed. 
image


image

I propose that a future issue be opened to add support for jumping back to the resources step. There does not seem to be any support for this from the patternfly side so it will take some thought re: how we want to go about this.

mturley:
@vconzola the Wizard component in PF today does not support automatically moving back to a step, unfortunately. If we want to be able to do that I can open an issue and get that fixed in PF. The only way to change the wizard step is via the back/next buttons and clicking on the steps in the left-side nav.

Maybe what we need to do is leave the nav visible on the results step, and then where we would have the "return to resources step" button we would need a message asking the user to go back using the nav. @ibolton336 proposed that we could have some kind of emphasis style on the step names/numbers in the nav which have issues (we can customize those). I don't see design guidelines for that, so we'd have to decide what to do there (different color? icon? bold text?)

If we do ideally want to be able to programmatically move back to another wizard step without requiring the user to click the step, let me know and I can open an issue in PatternFly for that. We'd have to add it as a future enhancement if we don't want to block this PR since it would require an upstream change.

mturley:
Actually... I don't love this idea, but if we really need to make a link back to another step without modifying PF, we could do it by programatically triggering a click event on that step in the nav (we could put a `ref` on that step's element and use it when the button/link is clicked). The nav would still need to be visible.

eriknelson:
I'd like to discuss this at our 2pm meeting with Vince today.

vconzola:
@ibolton336 [Mockups](https://docs.google.com/presentation/d/1ohU0yPW8ACtCbpxMf2FfvJ0deNzUT98xlFZ6Y5TBX94/edit?usp=sharing) updated per our discussion in Erik's meeting on Monday. (1) Updated popover text. (2) Don't try to give user an option to go back to the namespace selection step in the wizard.

ibolton336:
image

ibolton336:
image

ibolton336:
image

PR #808, opened 2020-04-15
Convert Storage class table (now Copy options) to PF, add claims and namespaces columns, other UX fixes

### LINKED ISSUE #587: Migration plan wizard - Step 4 enhancements
1. Table header prompt should be, "Select target storage class for copied PVs" - add the word "target" and remove the ":" at the end.
2. Use sentence style capitalization for navigation and table column headers.
3. Use PF dropdown component.
4. Us PF table component.

### LINKED ISSUE #773: Add "Claims" column to PVs table in storage class step in the mig plan wizard
@dymurray got feedback from a customer that PVC info would be useful at this step. We have room to add the column to the table.

### LINKED ISSUE #804: Bug 1823920 - Default storageclass in dropdown defaults to None instead of controller recommendation
Copied from: https://bugzilla.redhat.com/show_bug.cgi?id=1823920

Source cluster: mig-agnosticd 3.11 cluster
Targte cluster: mig-agnosticd 4.3.12 cluster (gp2 default storageclass)

As of master https://github.com/konveyor/mig-ui/commit/3ae13bb7c90b998019f118bc4769721cfae1e5aa

When selecting a namespace with a stateful workload and choosing copy for the pv action, the storageclass assignment screen currently defaults to displaying "None" instead of the controller specified default storageclass, as it did previously.

On the example clusters that I used, this results in a stage hanging during the restore process. The default value should have been gp2. This is a regression from previous behavior.

mturley:
Closes #587
Closes #773
Fixes #804

This PR renames the StorageClassForm and StorageClassTable to CopyOptionsForm and CopyOptionsTable, to match [Vince's mockups](https://docs.google.com/presentation/d/1hlP8wjzsO_9c9ri2rIz7o2kuIQvmKCNUhoQxTnUkHr0/edit#slide=id.g7f12813776_1_8). It adds Claim and Namespace columns, updates text to match the mockups, and converts the table to PF using the same sorting/filtering/pagination abstractions I wrote in https://github.com/konveyor/mig-ui/pull/776 and https://github.com/konveyor/mig-ui/pull/803.

In the process of rewriting the form and table components, I gave them the same treatment as the last 2 wizard steps (moved non-presentational logic out of the table component, removed redundant state, flattened unnecessary useEffect logic into render logic). While doing this I saw that the condition for setting recommended storage class settings would always be false (#804), so I fixed it in 6782c71 .

Since the copy options and storage class columns have values that don't come directly from the PV object (they perform lookups), I needed to enhance the `useFilterState` and `useSortState` hooks so they can take functions rather than string keys for determining the value to match/sort by. `useFilterState` now takes the `filterCategories` as an argument, and a filter category can optionally have a `getItemValue` function in it (if omitted, the category key will be used as before). The `useSortState` hook takes a `getSortValues` function instead of a `sortKeys` array.

I also changed the filter logic so rows will not match a filter if they have a null/undefined value for the selected field (so that "Target storage class: None" doesn't match every filter) and so that search filters are case insensitive.

This PR does not include the Verify copy column, that will be in a separate PR to address https://github.com/konveyor/mig-ui/issues/788.

# Before
Screenshot 2020-04-15 19 44 46

# After
Screenshot 2020-04-16 12 54 51

### DIFF
--- a/src/app/common/components/FilterToolbar/FilterToolbar.tsx
+++ b/src/app/common/components/FilterToolbar/FilterToolbar.tsx
@@ -28,6 +28,7 @@ export interface IBasicFilterCategory {
   key: string;
   title: string;
   type: FilterType; // If we want to support arbitrary filter types, this could be a React node that consumes context instead of an enum
+  getItemValue?: (item: any) => any;
 }
 
 export interface ISelectFilterCategory extends IBasicFilterCategory {
--- a/src/app/common/components/TableEmptyState.tsx
+++ b/src/app/common/components/TableEmptyState.tsx
@@ -1,0 +1,47 @@
+import React from 'react';
+import {
+  Flex,
+  FlexItem,
+  EmptyState,
+  EmptyStateVariant,
+  EmptyStateIcon,
+  EmptyStateIconProps,
+  Title,
+  EmptyStateBody,
+  Button,
+} from '@patternfly/react-core';
+import { SearchIcon } from '@patternfly/react-icons';
+import flex from '@patternfly/react-styles/css/utilities/Flex/flex';
+
+export interface ITableEmptyStateProps {
+  icon?: EmptyStateIconProps['icon'];
+  titleText?: string;
+  bodyText?: string;
+  onClearFiltersClick?: () => void;
+}
+
+const TableEmptyState: React.FunctionComponent<ITableEmptyStateProps> = ({
+  icon = SearchIcon,
+  titleText = 'No results found',
+  bodyText = 'No results match the filter criteria. Remove filters or clear all filters to show results.',
+  onClearFiltersClick,
+}: ITableEmptyStateProps) => (
+  <Flex className={flex.justifyContentCenter}>
+    <FlexItem>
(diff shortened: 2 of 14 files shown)


eriknelson:
> So what do you think about dropping the Migration type column (it's always Copy).

+1, it's entirely redundant

> Ideally, we skip the step entirely, but since we're tight on time, if it's easier to display an empty state for the table we should do that. I can provide text.

Also +1, I'd really like to explore the ability to conditionally skip wizard steps and see how expensive that will be to do with the current state of the code. For 1.2, let's go with an empty state (for some reason I feel like this had been implemented at one point?)

mturley:
Thanks @eriknelson @vconzola. Just pushed updates to address those comments.

Table with migration type column removed:
Screenshot 2020-04-16 12 54 51

Empty state (no results match the filters; I also added this one to the last 2 wizard steps):
Screenshot 2020-04-16 12 54 29

Empty state (no PVs selected for copy):
Screenshot 2020-04-16 12 54 08

PR #809, opened 2020-04-16
Plan name validation

gildub:
* Adds testName and nameError to utils
* Validate Plan name (Wizard) using testName

A Plan name is not a DNS name and can't allow double dashes.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1823830


gildub:
This is not addressing the initial issue, closing it.

PR #831, opened 2020-04-23
Yarn scripts: Add prettier:fix rule + prettier check to lint rule

gildub:
* Adds to yarn script lint prettier check for scss and json files
* Adds yarn scripts entry `prettier:fix` to formart scss and json files


package.json
```diff
@@ -12,8 +12,9 @@
     "image:build": "./scripts/build-image.sh",
     "image:push": "./scripts/push-image.sh",
     "clean": "yarn rimraf dist",
-    "lint": "yarn eslint . --ext .js,.jsx,.ts,.tsx",
-    "lint:fix": "yarn eslint . --ext .js,.jsx,.ts,.tsx --fix ",
+    "lint": "yarn eslint . --ext .js,.jsx,.ts,.tsx && yarn prettier ./src --check",
```
mturley, line 15: Won't `yarn prettier ./src --check` also run on all the JS/TS? With the eslint plugin, that would run prettier on those files twice.

package.json
```diff
@@ -12,8 +12,9 @@
...
+    "lint:fix": "yarn eslint . --ext .js,.jsx,.ts,.tsx --fix",
+    "prettier:fix": "yarn prettier ./src/ --write",
```
mturley, line 17: I would say if checking the SCSS formatting is part of `yarn lint`, fixing it should be part of `yarn lint:fix`. I'm fine with having `prettier:fix` be in here too in case you want to run just prettier without the rest of eslint while you're developing (it's faster), but we should probably add `&& yarn prettier:fix` to the end of `lint:fix`.

package.json
gildub: You're right, that was a test that I forgot to change. Fixed

package.json
gildub: Same as above I replaced that.

package.json
```diff
@@ -12,8 +12,9 @@
     "image:build": "./scripts/build-image.sh",
     "image:push": "./scripts/push-image.sh",
     "clean": "yarn rimraf dist",
-    "lint": "yarn eslint . --ext .js,.jsx,.ts,.tsx",
-    "lint:fix": "yarn eslint . --ext .js,.jsx,.ts,.tsx --fix ",
+    "lint": "yarn eslint . --ext .js,.jsx,.ts,.tsx && yarn prettier ./src/**/*.scss ./src/**/*.json --check",
+    "lint:fix": "yarn eslint . --ext .js,.jsx,.ts,.tsx --fix",
+    "prettier:fix": "yarn prettier ./src/**/*.scss ./src/**/*.json --write",
```
mturley, line 17: As mentioned on Slack, let's move this `yarn prettier [...] --write` command to be part of `lint:fix` and get rid of `prettier:fix`. If we do that and delete the unused `tslint.json` file I'll approve the PR

mturley: (and optionally we could have a `prettier:fix` that instead runs `yarn prettier . --write`. Although that would include .yaml files, which maybe we should add above?)

gildub: @mturley, that makes sense, I've added a commit to the patch, thanks.

PR #835, opened 2020-04-23
Unit tests for plan actions

gildub:
Using `__tests__` as place holder for tests files

Sort by alphabetical order PlanActionTypes because list is too long.

https://github.com/konveyor/mig-ui/issues/832


ibolton336:
@gildub Looks good. I am wondering though - where should we really be putting our effort in testing that would yeild the most bang for our buck? It doesn't seem like testing that the action types match up would give us a whole lot. Maybe we could write some test cases for some of our components that require user input i.e. the select components & form inputs. Then we could look into mocking out api calls & testing redux/saga operations for edit/delete/create functionality.

ibolton336:
The more I look into the react-testing-library @mturley mentioned, the more I like it. It would be a good fit for us since it focuses less on implementation details & more on "black box testing".  I also really like that it supports testing of some of the newer react features (hooks, context, etc) out of the box. 

https://npmcharts.com/compare/@testing-library/react,enzyme?interval=30

https://github.com/testing-library/react-testing-library#readme

PR #836, opened 2020-04-24
Stage Plan: disable 'cancel' when successful

gildub:
Closes https://github.com/konveyor/mig-ui/issues/834


eriknelson:
@gildub We need to get this updated by Friday morning if possible, can you take another look please?

gildub:
@eriknelson, this has been addressed, please review.

ibolton336:
> @eriknelson, this has been addressed, please review.

Looks good @gildub.

 @eriknelson after going through this with Gilles yesterday, we noticed that the derived state for table status appended to the migration object doesnt match up the current TS interface for MigMigration. Not sure what the best solution for remaining consistent with our TS support is here but wanted to give some background. @mturley Maybe you have an idea of the best way to do this?

gildub:
Thanks @ibolton336, I initially tried to augment  `src/client/resources/conversions.ts` but that didn't make sense and impacted sagas using it. 

So extending `IMigMigration` only at `MigrationActions` component level is minimal impact. Leaving further work to be done (in a separate patch?) with parent component `MigrationsTable`.

gildub:
Latest commit makes the extended type common to both `MigrationActions` and `MigrationsTable`.

src/app/home/components/DataList/Plans/MigrationsTable.tsx
```diff
@@ -132,6 +135,7 @@ const MigrationsTable: React.FunctionComponent<IProps> = ({ migrations, isPlanLo
     );
   }
 
+  console.log('This is migration', migrations);
```
mturley, line 138: Left in a console.log here 🙂

src/app/home/components/DataList/Plans/types.ts
```diff
@@ -0,0 +1,15 @@
+import { IMigMigration } from '../../../../../client/resources/conversions';
+
+export interface IMigration extends IMigMigration {
```
mturley, line 3: This makes perfect sense since tableStatus is extended in the selectors, and the extended version is what we pass around. I wonder if we should eventually make explicit interfaces for all of those types in `client/resources/conversions` though, I'm not sure how much the `ReturnType<typeof ...>` does for us (at least in some cases). I suppose since the params of `createMigMigration` have types and the returned object just reuses those directly, it's still useful. Might be worth looking at later.

gildub: @mturley, thanks for your review. 
I changed the interface name to `IMigrationWithStatus` it makes sense.

I was wondering about `ReturnType<typeof ...>` too. It seems handy but not as flex as formal interfaces. I'll leave it for another PR.

PR #840, opened 2020-04-24
Plan Wizard: Give focus to plan name

gildub:
Provides autofocus to input field after rendering using `useRef`.

### DIFF
--- a/src/app/plan/components/Wizard/GeneralForm.tsx
+++ b/src/app/plan/components/Wizard/GeneralForm.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
 import { FormikProps, FormikActions, FormikState } from 'formik';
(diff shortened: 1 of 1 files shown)


src/app/plan/components/Wizard/GeneralForm.tsx
```diff
@@ -24,10 +24,15 @@ const GeneralForm: React.FunctionComponent<IGeneralFormProps> = ({
   touched,
   values,
 }: IGeneralFormProps) => {
+  const inputRef = useRef(null);
   const onHandleChange = (val, e) => {
     handleChange(e);
   };
 
+  useEffect(() => {
```
mturley, line 32: As written, this will focus the input field every time this component renders, instead of just the first time. If we ever add any other fields here that could become problematic. If you want it to focus only after the first render you can pass an empty dependencies array as the second argument of `useEffect`. https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

Issue #663, opened 2020-02-11
UX improvements to the Add/Edit Cluster and Storage forms


eriknelson:
After our call with @vconzola and @mansam regarding custom CA bundle support, we decided to accept those PRs as they are since they encapsulate the functionality desired for the feature, and to separately address the UX improvements that Vince raised during the call so the two PRs are not conflated with one another. I'm filing this as a placeholder issue for now because I'd like to circle back with screenshots from master so we can itemize the UX changes required and get those fixed. As of now, I'm aware of:

* We need to ensure the correct buttons are marked as primary/secondary buttons, right now they are all primary.
* The "Is azure cluster prompt" should be above the URL, but below the name field.
* The close button should be a cancel button, as seen in the following mock provided by vince:

![1](https://user-images.githubusercontent.com/410933/74277573-c71cda00-4ce5-11ea-82cc-e3a9a212b95e.png)

![1-1](https://user-images.githubusercontent.com/410933/74277580-ca17ca80-4ce5-11ea-98cf-e822e13cb00c.png)

mturley:
Related to https://github.com/fusor/mig-ui/issues/356, discussing with @vconzola right now. I may fix this in the same PR.

eriknelson:
@mturley be aware we intend to bring in Sam's CA bundle functional changes to master prior to making these changes. You may want to implement this on top of his PR's commits.

mturley:
Gotcha, thanks for the heads up @eriknelson

vconzola:
Might also want to add a popover for SSL required field as part of this.

mturley:
Note to self: we should replace `react-select` with PatternFly's `Select` in the Storage form as part of this issue. Other occurrences of `react-select` are covered by https://github.com/konveyor/mig-ui/issues/736, https://github.com/konveyor/mig-ui/issues/585, https://github.com/konveyor/mig-ui/issues/586 and https://github.com/konveyor/mig-ui/issues/587.

mturley:
The CA bundle file field here in the cluster form should also be converted to a PF FileUpload component as part of this issue.

@vconzola , this is lower priority, but you suggested a popover on the "Require SSL verification" field here. Do you know what text that should have?

vconzola:
@mansam @eriknelson Can you provide some text explaining when a user would select the "Require SSL verification" checkbox?

mturley:
When you guys get a chance to figure out the text for the "Require SSL verification" help info, I plan to get this done asap.

[...]

PR #841, opened 2020-04-27
Split AWS storage form into separate "S3" and "AWS S3" forms, add SSL helper text on cluster and storage forms

### LINKED ISSUE #779: Need to help users understand when they should provide S3 endpoint and when to ignore
Trying to be careful to avoid jumping to an implementation, I've seen a lot of users guess at an AWS base endpoint to hit. If they are using AWS, they should *not* input anything for an S3 root endpoint, because this is the default.

The obvious, immediate way to convey that is a tooltip or better: a note visible by default that says the above. I think that's temporary duct tape, so I wanted to file this to take a step back and think about that UX.

### LINKED ISSUE #829: Determine default SSL verify state and add hint text
The current default behavior for adding a cluster has the SSL verify checkbox
checked. This is going to lead to a LOT of failed connections if people skip
the option because they don't know what it does, and they're using a self-signed
cert for their cluster.

We should think about what the default state for this checkbox should be for
both clusters and storage, update the code to reflect that, and add hint text
that steers the user in the right direction so they know when they should check
the box, and when they will need to upload their CA.

mturley:
Closes #779
Closes #829 

* Renames the existing AWSForm component to S3Form
* Adds a boolean prop `isAWS` to that component, conditionally renders the S3 endpoint field (now required) and SSL-related fields only when `isAWS` is false, and hard-codes the requireSSL field to true when `isAWS` is true.
* Splits the "S3" select option into two options "AWS S3" and "S3", which both render the S3Form with `isAWS` true and false respectively.
* Adds a new annotation on the MigStorage object to persist which of those two options was selected, so it can be correctly pre-populated when reopening the form for editing (with fallback logic to guess the right one for pre-existing storages that don't have the annotation)
* Adds TypeScript types for the components and MigStorage objects being passed around here, fixes a few bad references revealed by that
* Adds helper text to the SSL checkbox fields in S3Form and AddEditClusterForm (text from https://github.com/konveyor/mig-ui/issues/663#issuecomment-619101758)

The rest of the items in #663 will be addressed in a separate PR.

# Screens

## Before

Screenshot 2020-04-29 12 33 52

Screenshot 2020-04-29 12 33 37

## After

Screenshot 2020-04-29 12 34 34

Screenshot 2020-04-29 12 34 46

Screenshot 2020-04-29 12 35 00

### DIFF
--- a/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
+++ b/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
@@ -194,7 +194,16 @@ const InnerAddEditClusterForm = (props: IOtherProps & FormikProps<IFormValues>) 
           />
         </FormGroup>
       )}
-      <FormGroup fieldId={requireSSLKey}>
+      <FormGroup
+        fieldId={requireSSLKey}
+        helperText={
+          <>
+            Select 'Require SSL verification' to verify SSL connections to the cluster. If you are
+            using a self-signed certificate, you may either disable SSL verification, or upload your
+            certificate bundle.
+          </>
+        }
+      >
         <Checkbox
           onChange={formikHandleChange}
           onInput={formikSetFieldTouched(requireSSLKey)}
--- a/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
+++ b/src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
@@ -308,7 +308,7 @@ const AddEditClusterForm: any = withFormik({
       token: '',
       isAzure: false,
       azureResourceGroup: '',
-      requireSSL: true,
+      requireSSL: false,
       caBundle: '',
     };
 
--- a/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
+++ b/src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
@@ -16,6 +16,11 @@ interface IOtherProps {
   storageList: IStorage[];
 }
 
+interface IProviderOption {
+  label: string;
+  value: 'aws-s3' | 'generic-s3' | 'gcp' | 'azure';
+}
+
 const AddEditStorageForm = (props: IOtherProps) => {
   const { storageList } = props;
   const storageContext = useContext(StorageContext);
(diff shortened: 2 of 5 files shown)


src/app/cluster/components/AddEditClusterModal/AddEditClusterForm.tsx
```diff
@@ -194,7 +194,16 @@ const InnerAddEditClusterForm = (props: IOtherProps & FormikProps<IFormValues>)
           />
         </FormGroup>
       )}
-      <FormGroup fieldId={requireSSLKey}>
+      <FormGroup
+        fieldId={requireSSLKey}
+        helperText={
+          <>
+            Select 'Require SSL verification' to verify SSL connections to the cluster. If you are
```
eriknelson, line 201: +1

src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
```diff
@@ -1,18 +1,24 @@
...
   onClose: any;
   addEditStatus: any;
   checkConnection: (name) => void;
-  storageList: any;
+  storageList: IStorage[];
+}
+
+interface IProviderOption {
+  label: string;
+  value: 'aws-s3' | 'generic-s3' | 'gcp' | 'azure';
```
eriknelson, line 21: Interesting, didn't know this was possible.

mturley: Yep, union types. I'm using them here instead of an enum just for brevity, but they can also be used for polymorphism since you can union values of multiple types

mturley:
> can we make the default unchecked for the Cluster Add screen?

sure thing.

src/app/storage/components/AddEditStorageModal/AddEditStorageForm.tsx
mturley: Interesting stuff I was reading on this: https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant

jwmatthews:
This looks great, chatted with @tsanders-rh on this and both us like what this achieves, looks good.

+1 from us

mturley:
Thanks for merging @eriknelson! You had perfect timing, I was just putting the finishing touches on https://github.com/konveyor/mig-ui/pull/846. So that's rebased now and open for review.

PR #842, opened 2020-04-27
Bug 1818021 - Implement observedDigest

### LINKED ISSUE #781: Re-running PV discovery shows wrong info when navigating back & changing namespaces
This issue stems from comparing the generation fields on the currentPlan & the response from a getPlan request when polling. There are cases when the generation is stale or not behaving as we expected making it impossible to determine when to rerun pv discovery/when to timeout pv discovery/etc

Related #777

ibolton336:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1818021

Closes #781

### DIFF
--- a/src/app/plan/duck/sagas.ts
+++ b/src/app/plan/duck/sagas.ts
@@ -273,6 +273,7 @@ function* checkUpdatedPVs(action) {
   let pvUpdateComplete = false;
   let tries = 0;
   const TicksUntilTimeout = 240;
+  const { isRerunPVDiscovery } = action;
   while (!pvUpdateComplete) {
     if (tries < TicksUntilTimeout) {
       tries += 1;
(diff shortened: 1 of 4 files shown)


src/app/plan/duck/sagas.ts
```diff
@@ -287,14 +287,20 @@ function* checkUpdatedPVs(action) {
...
-          const oldGeneration = currentPlan.metadata.generation;
-
-          //Generation check incremented twice: once for ui change, once for controller change.
-          if (isRerunPVDiscovery) {
-            return updatedGeneration >= oldGeneration + 2;
+          if (updatedPlan.status) {
+            if (isRerunPVDiscovery) {
+              const updatedGeneration = updatedPlan.status.observedDigest;
+              const oldGeneration = currentPlan.status.observedDigest;
+              if (updatedGeneration === oldGeneration) {
```
eriknelson, line 294: Can we update each of these return paths with a comment that describes the specific case?

eriknelson:
I've tested this on 3->3 and 3->4 migrations and had good success with both, this even repairs a number of edit problems we've had previously:

ObservedDigestSuccess

Really happy to see this come together, thank you @ibolton336!

Want to follow up on the comments because I'm having a hard time following but I think this is looking good.

PR #843, opened 2020-04-28
Update DNS1123 error message to be object-agnostic

gildub:
Error message must be generic as it's used by several components

invalid-plan

### DIFF
--- a/src/app/common/duck/utils.ts
+++ b/src/app/common/duck/utils.ts
@@ -40,9 +40,9 @@ export const handleSelfSignedCertError = (failedUrl: string, dispatch: any) => {
 };
 
 const DNS1123Error = (value) => {
-  return `Invalid character: "${value}". Plan names must be DNS-1123 label compliant, starting and
-    ending with a lowercase alphanumeric character and containing only lowercase alphanumeric
-    characters, "." or "-".`;
+  return `Invalid character: "${value}". Name must be DNS-1123 label compliant, starting and ending
+    with a lowercase alphanumeric character and containing only lowercase alphanumeric characters, "."
+    or "-".`;
 };
 
 const testDNS1123 = (value) => DNS1123Validator.test(value);


eriknelson:
To clarify, `MigPlan` names *are* k8s DNS names and the API server allows the creation of a `MigPlan` with a double dash in the name. Due to the fact that an `ImageStream` is spawned during `MigPlan` reconciliation using the name, it fails reconcilation. For some reason, OCP Image related objects **fail** validation, including `ImageStream` objects, when a double dash is used in the name. This is inconsistent with general k8s validation, I plan to file a bz against OCP.

src/app/common/duck/utils.ts
```diff
@@ -40,13 +41,20 @@ export const handleSelfSignedCertError = (failedUrl: string, dispatch: any) => {
 };
 
 const DNS1123Error = (value) => {
-  return `Invalid character: "${value}". Plan names must be DNS-1123 label compliant, starting and
-    ending with a lowercase alphanumeric character and containing only lowercase alphanumeric
-    characters, "." or "-".`;
+  return `Invalid character: "${value}". Must be DNS-1123 label compliant, starting and ending with
```
eriknelson, line 44: I don't think this language change is necessary here.

src/app/common/duck/utils.ts
```diff
@@ -40,13 +41,20 @@ export const handleSelfSignedCertError = (failedUrl: string, dispatch: any) => {
...
+};
+
+const nameError = (value) => {
+  return `Invalid name: "${value}". Plan names must start and end with a lowercase alphanumeric
+  character and can contain only lowercase alphanumeric character or one subsequent '-'`;
 };
 
 const testDNS1123 = (value) => DNS1123Validator.test(value);
 
+const testName = (value) => NameValidator.test(value);
```
eriknelson, line 56: We should have a plan name specific function that first uses the testDNS1123 validation, and if it passes, then additionally applies a validation that ensures no consecutive double dashes are used. Keeps a clean separation of concerns and we can comment specifically the reason for the double dash validation.

src/app/common/duck/utils.ts
gildub: Actually `DNS1123Error` isn't specific to Plan name. The error message should be more generic. That was changed in https://github.com/konveyor/mig-ui/pull/814 (by me but I missed that).

eriknelson:
Sorry @gildub, I spoke too soon about requiring this. Some new info arose from @sseago that the controller is actually smart enough to deal with this on the backend and it uses UIDs for ImageStreams, we've confirmed that's functional and feel we should accept any plan name that will pass the DNS1123 test as that's the k8s test. Let's maybe use this PR to update the error message and rename, and remove any logic that rejects the double dashes?

gildub:
@vconzola, please check the error message change.

eriknelson:
Will wait to merge this until we get an ack from Vince

vconzola:
looks good to me.

PR #852, opened 2020-05-04
Render AppComponent: use body as root element

gildub:


ibolton336:
Going to close this one. If there is an issue this is solving we should address it again in the future.

PR #856, opened 2020-05-07
Clean up pv discovery logic

### LINKED ISSUE #845: Refactor PV discovery status poll logic
Currently, the pv discovery condition is hard to follow. This is an issue to track the refactoring.

Action items:
- Make redux action descriptions & matching function names on the UI side consistent for readibility
- extract pv discovery logic into its own saga function that gets triggered by the update plan function rather than getting mixed in with
- Have solid definitions of all cases that trigger pv discovery to run.

ibolton336:
Closes #845


src/app/plan/duck/sagas.ts
```diff
@@ -120,9 +120,7 @@ function* planPatchClose(planValues) {
       closedPlanSpecObj
     );
     yield put(PlanActions.updatePlanList(patchPlanResponse.data));
-    yield put(PlanActions.planUpdateSuccess());
```
eriknelson, line 123: Is this intended?

ibolton336: Yep - there is no more plan update action.

PR #865, opened 2020-05-13
Bug 1835277: fix encryption regression for caBundle field value

ibolton336:
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1835277


ibolton336:
Ah apparently if you force push a pr back to master it auto closes.

PR #875, opened 2020-05-21
Lazy loading api support

### LINKED ISSUE #659: Namespace pages table needs to correctly use the paged discovery API
The discovery service has a supported paging API, while the current implementation of the namespace table actually loads all of the pages at once rather than lazily loading the relevant page the current user has selected. We'll want to hook that up so next/previous loads in the missing chunk of namespaces.

FYI @jortel, desire is to get this hooked up shortly.

ibolton336:
Closes #659


ibolton336:
Closing for now. Will reopen when this is prioritized.

PR #879, opened 2020-05-27
Add check for existing plan name

### LINKED ISSUE #839: Add plan wizard should reject one if it already exists
Just accidentally tried to create a plan that used the same name as one that
already exists. The wizard allows me to continue to the PV discovery step,
at which point it just toasts and says plan name failed (backend server probably
error'd on the POST). We should gracefully handle this validating the original
name before continuing to the next step of the wizard with a backend call that
ensures it doesn't already exist.

ibolton336:
Closes #839


ibolton336:
@vconzola

eriknelson:
@vconzola I need to merge this to get a release cut, please file an issue if you would like that copy to change.

PR #910, opened 2020-06-08
Use generateName for MigStorage Secrets

### LINKED ISSUE #908: generateName should be used for migstorage secrets in addition to clusters
The secret that gets created associated with a MigCluster uses `generateName`
to ensure a random name is used, and the ref is set on the MigCluster.

The same behavior should be used for MigStorage associated secrets, which
is not currently implemented.

pranavgaikwad:
Fixes #908 

Uses `generateName` instead of `name` for Secret referenced by MigStorage.

### DIFF
--- a/src/app/storage/duck/sagas.ts
+++ b/src/app/storage/duck/sagas.ts
@@ -8,6 +8,7 @@ import {
   createMigStorage,
   updateMigStorage,
   updateStorageSecret,
+  getTokenSecretLabelSelector,
 } from '../../../client/resources/conversions';
 import { StorageActions, StorageActionTypes } from './actions';
 import { AlertActions } from '../../common/duck/actions';
@@ -85,8 +86,18 @@ function* removeStorageSaga(action) {
     );
     const migStorageResource = new MigResource(MigResourceKind.MigStorage, migMeta.namespace);
 
-    const arr = yield Promise.all([
-      client.delete(secretResource, name),
+    const secretResourceList = yield client.list(
+      secretResource,
+      getTokenSecretLabelSelector(MigResourceKind.MigStorage, name)
+    );
+
+    const secretResourceName =
+      secretResourceList.data.items && secretResourceList.data.items.length > 0
+        ? secretResourceList.data.items[0].metadata.name
+        : '';
+
+    yield Promise.all([
+      client.delete(secretResource, secretResourceName),
       client.delete(migStorageResource, name),
     ]);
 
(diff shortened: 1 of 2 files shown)


src/app/storage/duck/sagas.ts
```diff
@@ -85,8 +86,18 @@ function* removeStorageSaga(action) {
...
-      client.delete(secretResource, name),
+    const secretResourceList = yield client.list(
+      secretResource,
+      getTokenSecretLabelSelector(MigResourceKind.MigStorage, name)
+    );
+
+    const secretResourceName =
+      secretResourceList.data.items && secretResourceList.data.items.length > 0
+        ? secretResourceList.data.items[0].metadata.name
+        : '';
```
eriknelson, line 97: This is an odd code branch. Do we want to ask the client to delete with a resource name of `''`? It seems like this is an unexpected result.

src/app/storage/duck/sagas.ts
```diff
@@ -170,18 +193,36 @@ function* addStorageRequest(action) {
...
+      });
+
+      Object.assign(migStorage.spec.volumeSnapshotConfig.credsSecretRef, {
+        name: storageSecretAddResult.data.metadata.name,
+        namespace: storageSecretAddResult.data.metadata.namespace,
+      });
+
+      const storageAddResult = yield client.create(migStorageResource, migStorage);
+
+      if (storageAddResult.status === 201) {
```
eriknelson, line 214: I don't think this should only get pushed to the results array if it was successfully created. We want a failed result to trigger the rollback?

src/app/storage/duck/sagas.ts
```diff
@@ -192,7 +233,7 @@ function* addStorageRequest(action) {
       // The objects that need to be rolled back are those that were fulfilled
       const rollbackObjs = storageAddResults.reduce((rollbackAccum, res) => {
         return res.state === 'fulfilled'
```
eriknelson, line 235: Will this still work now that you're pushing actual results into the list instead of a promise?

src/app/storage/duck/sagas.ts
pranavgaikwad: @eriknelson This will work. Notice changes on line 224 where I have changed the logic to look for actual response status code, instead of the promise' fulfilled or rejected status

src/app/storage/duck/sagas.ts
pranavgaikwad: @eriknelson Good catch. totally forgot about this piece. will make changes

src/app/storage/duck/sagas.ts
pranavgaikwad: @eriknelson So I assumed when we create a cluster, it either gets created or it doesn't. I did not think that there's a half way outcome. According to that logic, I thought it would be redundant to add it to rollback objects.

src/app/storage/duck/sagas.ts
pranavgaikwad: @eriknelson Agreed. However, I did not have a good answer to this question : What if Secret does not exist at all while we are deleting the Storage? Removing a blank resource doesn't have an adverse effect on the outcome. I agree that it looks like an odd branch. Open for better idea to handle this. I just wanted to leverage benefits of Promise.all() and make less changes to current code.

eriknelson:
@pranavgaikwad FYI we merged another big PR to master, need to rebase this.

src/app/storage/duck/sagas.ts
```diff
@@ -136,92 +145,57 @@ function* addStorageRequest(action) {
...
+
+      Object.assign(migStorage.spec.volumeSnapshotConfig.credsSecretRef, {
+        name: createdStorageSecret.metadata.name,
+        namespace: createdStorageSecret.metadata.namespace,
+      });
+
+      storageAddResults.push(yield client.create(migStorageResource, migStorage));
+    } catch {
+      // attempt rolling back secret
+      yield client.delete(secretResource, createdStorageSecret.metadata.name);
```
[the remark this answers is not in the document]
eriknelson: I don't think it's redundant in that it's not doing useful work. This ensures we aren't just leaking secrets.

PR #899, opened 2020-06-02
Add new (incomplete) Tokens page

mturley:
Working towards https://github.com/konveyor/mig-ui/issues/885.

This PR adds a new TokensPage and TokensTable, with a new directory at src/app/token/duck for types/reducers/etc and a new route and navigation item. I added a reducer and some stub action creators as a basis for integrating with the k8s client, but we'll need to add more actions/handlers as we get it working end-to-end. The page is connected to Redux but currently only pulls in the clusterList (the tokenList is mock/fake data so far). I have also not added an add/edit modal yet, I plan to do so in the next PR. The type interface for MigToken is correct (based on the latest CRD) and fields in the table are properly formatted from this structure. Dropping in real data should "just work".

Remaining work to close #885 after this: finish CRUD redux code, connect to API via k8s client changes from @eriknelson, implement Add/Edit modal with complete OAuth flow.

Screenshot 2020-06-08 16 43 54

Issue #892, opened 2020-06-01
Implement add token for cluster


eriknelson:
WIP

When "Add Token" modal is opened from the Clusters page, default to type=OAuth

* Associate with cluster on Add Token modal -- if adding a token from the cluster
page, should that default to the contextual cluster and be a disabled dropdown,
but should be *enabled* if the modal was launched from the Tokens page?

Selecting "Generate OAuth Token" button should redirect the user to the OAuth
page of the cluster that was specified in the drop down. The user logs in,
and the OAuth server should redirect the user back to the UI.

Non-admin users should also have the opportunity to create a MigToken with an
SA token to be used for authentication purposes.

Issue #885, opened 2020-06-01
Implement tokens screen


eriknelson:
Need to implement token management screen with a link in the left nav of the
application.

Table should have a type column indicating whether it is an OAuth or SA type
(this should be made available after the token has been created by the controller
annotating this information on the `MigToken` object inside `status`). It will
list the associated cluster, as well as the expiration (also set by the controller
in the status field).

The tokens each have a "Status" that indicates if its Expired, Expiring soon,
or ok.

Expiring soon threshold should be easy to change, though probably doesn't require a
configmap value. Maybe 2 hours to start with.

mturley:
This was auto-closed by merging #899 because of my poor phrasing of "Remaining work to close #885 after this..." 😆 reopening since I imagine we only want to close this when it's working end-to-end with the API, right @eriknelson ?

eriknelson:
+1,its kind of broad in scope maybe we can come back to this and add some sub tasks. It should be open. Funny I saw your PR and figured it wouldn't actually close this one but I guess we have our answer

PR #911, opened 2020-06-09
Implement "Add Token" modal on Tokens page with stubs and OAuth flow using localStorage

mturley:
Part 2 of #885 (part 1 was #899). Adds components that will be reused for #892.

Adds the AddEditTokenModal and AddEditTokenForm to src/app/common/components so they can be used in multiple pages. For now they are only used on the Tokens page. Formik is set up for the fields and validation, and the modal is connected to Redux, but no actions are wired up.

Fields are in place based on [this mockup](https://docs.google.com/presentation/d/1JU_Lw1A-gUO4gBRofZzi74aONcknYRHbqBl_xHxikYQ/edit#slide=id.g8803b278a4_8_43):

Screenshot 2020-06-10 17 54 41

I also added a proof of concept for the OAuth login flow using a self-closing new tab for the login page. It works very well, and has the benefit of preserving all of the original tab's state without any difficult-to-maintain state persistence code.

To test the redirect/storage approach, I stood up a fake cluster login page at https://cam-oauth-testing-login.surge.sh/. That page will take a URL parameter called `state` and when you click the login button it will redirect you to `http://localhost:9000/oauth-landing?state=${state}&code=mock-code-12345` (with a random code each time). So it is useful for making sure the landing page stores that mock authorization code in the correct place and the Add token modal reacts properly to that event. I use the `state` param to uniquely identify each login attempt so the storage event listener is only looking for that specific response (in the unlikely event that the user is trying to do this process in more than one window at the same time, it will still work).

Ooxy4F0OmC

That new tab closes itself immediately after reaching my new `/oauth-landing` page, but if for some reason it fails to close itself (possible user-specified browser restriction? not sure if that's a thing) this is what that landing page looks like:

Screenshot 2020-06-10 18 55 42


package.json
```diff
@@ -93,6 +93,7 @@
     "reduxsauce": "^1.0.1",
     "reselect": "^4.0.0",
     "sass": "^1.22.10",
+    "shortid": "^2.2.15",
```
mturley, line 96: Not sure if you guys think this is necessary / worth the additional little dependency. I just use shortid to generate a little unique localStorage key so that in theory if the user had two OAuth login attempts going at once they wouldn't collide.

eriknelson: Seems fine to me, I think I've used this in other projects previously.

mturley:
Rebasing now to resolve conflicts, I'll push again in a few.

src/app/home/HomeComponent.tsx
```diff
@@ -1,22 +1,12 @@
...
-  Brand,
-} from '@patternfly/react-core';
+import { Nav, NavList, NavItem, Page, PageSidebar, SkipToContent } from '@patternfly/react-core';
 import { PollingContext } from '../home/duck/context';
 import { ClustersPage, StoragesPage, PlansPage, LogsPage, TokensPage } from './pages';
 import RefreshRoute from '../auth/RefreshRoute';
-import openshiftLogo from '../common/components/CAM_LOGO.svg';
 import { ICluster } from '../cluster/duck/types';
-const styles = require('./HomeComponent.module');
+import PageHeaderComponent from '../common/components/PageHeaderComponent';
```
eriknelson, line 9: +1

src/app/common/components/AddEditTokenModal/AddEditTokenForm.tsx
```diff
@@ -0,0 +1,256 @@
...
+          isChecked={values.tokenType === TokenType.oAuth}
+          onChange={(checked) => {
+            if (checked) setFieldValue(FieldKey.tokenType, TokenType.oAuth);
+          }}
+          label="OAuth"
+        />
+        <div className={`${spacing.mtMd} ${spacing.mbLg} ${spacing.mlLg}`}>
+          <TextContent>
+            <Text component="p">
+              Log in to the cluster to automatically generate the token, or enter the token string
```
eriknelson, line 155: We aren't going to accept user provided OAuth tokens, so I think the copy after the comma should be removed. @vconzola thoughts on that?

eriknelson: Not going to hold up this PR on this since this is all under construction.

mturley: Maybe that's referring to the Service Account field below, and this help text should be above the radio button choice (if it describes both)?

vconzola: @mturley Nope, @eriknelson is correct. That text is leftover from when I thought user could enter an Oauth token string. It's removed from my latest mockups.

mturley: Ah, ok thanks @vconzola. When we get this all connected we'll have to do a once-over of your mockups to make sure we got all these little details.

mturley: Are these still your latest mockups? I still see this text in there: https://docs.google.com/presentation/d/1JU_Lw1A-gUO4gBRofZzi74aONcknYRHbqBl_xHxikYQ/edit#slide=id.g801d08d2ef_0_16

vconzola: @mturley Dang! I must have updated the mockup in Balsamiq but never copied it into Slides. Will do that this morning.

vconzola: @mturley Mockups updated with new text. https://docs.google.com/presentation/d/1JU_Lw1A-gUO4gBRofZzi74aONcknYRHbqBl_xHxikYQ/edit?usp=sharing

PR #912, opened 2020-06-11
[WIP] Add OAuth login flow using a localStorage listener and a self-closing new tab

mturley:
This PR is currently branched from #911 and includes those commits. The changes unique to this PR are in the last 3 commits. See that diff here: https://github.com/mturley/mig-ui/compare/add-token-modal...mturley:tmp-oauth-flow

WIP. This serves as a proof of concept for the OAuth login flow using a self-closing new tab for the login page. It works very well, and has the benefit of preserving all of the original tab's state without any difficult-to-maintain state persistence code.

To test the redirect/storage approach, I stood up a fake cluster login page at https://cam-oauth-testing-login.surge.sh/. That page will take a URL parameter called `state` and when you click the login button it will redirect you to `http://localhost:9000/oauth-landing?state=${state}&code=mock-code-12345` (with a random code each time). So it is useful for making sure the landing page stores that mock authorization code in the correct place and the Add token modal reacts properly to that event. I use the `state` param to uniquely identify each login attempt so the storage event listener is only looking for that specific response (in the unlikely event that the user is trying to do this process in more than one window at the same time, it will still work).

Ooxy4F0OmC

That new tab closes itself immediately after reaching my new `/oauth-landing` page, but if for some reason it fails to close itself (possible user-specified browser restriction? not sure if that's a thing) this is what that landing page looks like:

Screenshot 2020-06-10 18 55 42


mturley:
This PR is being combined with https://github.com/konveyor/mig-ui/pull/911, closing.

PR #914, opened 2020-06-15
Non-admin: changes to General step of plan wizard

### LINKED ISSUE #889: Implement non-admin MigPlan General step
WIP

This issue spans dedicating the general step for cluster name, and adding
cluster src, dest designation, in addition to adding the new token assignment.

It's likely this will spawn other issues as the mocks continue to evolve to
be more of a guided flow that allows the user to add things like tokens to their
migration if they haven't already added one.

mturley:
Closes #889.

This PR moves the cluster and storage selections from the ResourceSelectForm into GeneralForm, renames the ResourceSelectForm to NamespacesForm, adds "associated token" fields under each cluster field in a grid layout, and provides a control to open the AddEditTokenModal from the token dropdowns when they are empty. If there is only one token available for a selected cluster, it is pre-selected when the cluster is selected. When creating a new token, the selected cluster is pre-selected in the add token modal.

The expiration status of the selected token is pulled from the existing `getTokenInfo` helper from #899, so error/warning icons are shown next to each expired/expiring token in the dropdowns, and when one of those tokens is selected you get a message below the dropdown offering to let you regenerate it (that link is currently a stub).

I rebased this on top of #919, and I'm pretty sure if you add a token from the dropdown here it it will successfully come back and be selected (although I haven't been able to test this on my cluster yet). When that happens, you get a happy green checkmark under the dropdown saying the token has been associated. If you change the source or target cluster selection after any of these messages appear under the corresponding token dropdown, the message is hidden. There was a lot of repetition for all this state in each of the token dropdowns, so I abstracted them into a reusable TokenSelect component that is fairly standalone in case we ever need to select a token somewhere else in the UI.

When building this grid layout, I was careful to preserve the order of fields (source cluster, associated token, target cluster, associated token) when the UI collapses into mobile-responsive mode (each cluster/token field pair is one grid item).

# Screens

## Layout
Screenshot 2020-06-17 21 14 52

## Token selection
Screenshot 2020-06-17 21 15 18

## Expired message
Screenshot 2020-06-17 21 15 27

(same thing can appear with an orange triangle and an "expiring soon" message, but I didn't capture a screenshot of that).

## No tokens found, opportunity to add one
Screenshot 2020-06-17 21 15 43

## Adding a token from within the wizard
The cluster you already chose is pre-selected and that dropdown is disabled.
Screenshot 2020-06-17 21 38 32

## Inline confirmation after new token is added/associated
Screenshot 2020-06-17 21 17 12

## Mobile layout
Screenshot 2020-06-17 14 53 55


mturley:
This PR does not implement the inline alert message that describes the newly generated OAuth token (from [this mockup](https://docs.google.com/presentation/d/1JU_Lw1A-gUO4gBRofZzi74aONcknYRHbqBl_xHxikYQ/edit#slide=id.g89a48df6ed_0_34)), I think that may be out of scope here, let's loop back on that when we have the OAuth flow working end-to-end.

Screenshot 2020-06-17 21 43 20

Issue #917, opened 2020-06-17
Action Creators should correctly type their params


eriknelson:
Many of the Action Creators are using bad types for their parameters, imported
from ReturnType of the conversion functions. This is incorrect, and instead
they should be using the I<Model> variants that are the composite objects that
hold both the Mig object + its references (like Secrets). Usually these types
are found in a types.ts file, although we are inconsistent.

A PR sufficient for closing this issue should ensure that the described types
all exist, follow the same name convention, and that the action creator's
arguments are correctly typed to accept them

`Ex: updateCusters: ICluster[] NOT updateClusters: IMigCluster[]`

I've just been burned for a few hours trying to debug an issue that would have
been caught immediately by the compiler had I typed this correctly :)

PR #919, opened 2020-06-17
Adds initial support to connect to token backend

### LINKED ISSUE #885: Implement tokens screen

### LINKED ISSUE #891: Add MigToken support to k8s client
Need to add MigToken support to the CAM UI's k8s client so we have API access
similar to the rest of the objects.

eriknelson:
This PR is another WIP PR adding incremental non-admin feature support; it includes:
* Adds MigToken support to the client
* Adds token polling
* Hooks up the Tokens table to the redux tree and removes fake tokens
* Adds "Add Token" support for SAs
* Some amount of refactoring where it made sense (ex: moving the token form values into the token types, since it is used in at least 2 places other than the form component itself)

It does **not** include:
* Token edit/delete
* OAuth token support (will follow up with this)

I also discovered #917 while implementing this but refrained from fixing it and adding scope creep. I'd like to follow up this PR with a distinct PR to repair this issue as well.

Closes #891
Closes #885

### DIFF
--- a/src/app/token/duck/actions.ts
+++ b/src/app/token/duck/actions.ts
@@ -1,6 +1,4 @@
-// NATODO do we need all of these? figure out the CRUD/polling
-
-import { IToken } from './types';
+import { IToken, ITokenFormValues } from './types';
 
 export const TokenActionTypes = {
   MIG_TOKEN_FETCH_REQUEST: 'MIG_TOKEN_FETCH_REQUEST',
@@ -22,6 +20,25 @@ export const TokenActionTypes = {
 };
 
 // NATODO what can we abstract out here? Take lessons from Migration Analytics?
+const addTokenRequest = (tokenValues: ITokenFormValues) => ({
+  type: TokenActionTypes.ADD_TOKEN_REQUEST,
+  tokenValues,
+});
+
+const addTokenSuccess = (newMigToken: IToken) => ({
+  type: TokenActionTypes.ADD_TOKEN_SUCCESS,
+  newMigToken,
+});
+
+const addTokenFailure = (error) => ({
+  type: TokenActionTypes.ADD_TOKEN_FAILURE,
+  error,
+});
+
+const updateTokens = (updatedTokens: IToken[]) => ({
+  type: TokenActionTypes.UPDATE_TOKENS,
+  updatedTokens,
+});
 
 const migTokenFetchRequest = () => ({
   type: TokenActionTypes.MIG_TOKEN_FETCH_REQUEST,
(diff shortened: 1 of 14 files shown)


src/client/resources/conversions.ts
```diff
@@ -682,6 +682,50 @@ export function createMigHook(migHook: any, namespace: string) {
...
+  namespace: string,
+  migTokenSecretName: string,
+  migTokenSecretNamespace: string,
+  migClusterName: string
+) {
+  return {
+    apiVersion: 'migration.openshift.io/v1alpha1',
+    kind: 'MigToken',
+    metadata: {
+      name,
```
ibolton336, line 713: should we use the generateName field for the token name too?

src/app/token/duck/sagas.ts
```diff
@@ -1,2 +1,156 @@
...
+
+function* watchAddTokenRequest() {
+  yield takeLatest(TokenActionTypes.ADD_TOKEN_REQUEST, addTokenRequest);
+}
+
+export default {
+  // NATODO: Implement and/or remove unecessary copies
+  // watchRemoveClusterRequest,
+  watchAddTokenRequest,
+  // watchUpdateClusterRequest,
```
ibolton336, line 153: cleanup

src/app/token/duck/actions.ts
```diff
@@ -36,4 +53,33 @@ const migTokenFetchFailure = () => ({
...
+  params,
+});
+
+const stopTokenPolling = () => ({
+  type: TokenActionTypes.TOKEN_POLL_STOP,
+});
+
+// NATODO: Implement and/or remove unecessary copies
+export const TokenActions = {
+  // removeClusterRequest,
```
ibolton336, line 67: cleanup

src/client/resources/conversions.ts
eriknelson: Don't think so. It's a primary Mig object like a MigCluster or MigStorage object and the user provides a name. We've discussed maybe decoupling the name the user provides from the actual name that's used on the object so it doesn't have to abide by kubernetes naming requirements as a primary key, but I think that's distinctly separate from this.

vconzola:
@ibolton336 That's correct. We should use an empty state component inside of table as shown here: https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/empty-state

eriknelson:
Added empty state for tokens:
image

mturley:
Hey, sorry I didn't look at this sooner. I had originally left out the empty state for tokens because Vince told me to, since we assumed the user would always have at least the one token they logged in with.

vconzola:
Sure @mturley, blame it on the designer who doesn't know anything.  😉  @eriknelson didn't you tell me that users will always have at least one token? That's what I have in my notes because "user can't delete the host cluster token".

eriknelson:
Yeah I forgot about that, I don't expect the host token to ever be invalid or to be removable. It's exposed here because we don't have the auto creation of the logged in token implemented yet.

mturley:
Haha, sorry Vince. I was pretty confident you were right, if that helps 🙃

PR #924, opened 2020-06-18
Prevent disabled add plan button from suppressing tooltip

mturley:
At some point, something about the `<Tooltip>` behavior changed and it no longer works when the immediate child is disabled. So when the user has no plans, the "Add migration plan" button is disabled with no indication of why. Wrapping the button in a div fixes the hover event somehow, so this tooltip works again.

Screenshot 2020-06-18 13 44 19

### DIFF
--- a/src/app/home/pages/PlansPage/PlansPage.tsx
+++ b/src/app/home/pages/PlansPage/PlansPage.tsx
@@ -119,15 +119,19 @@ const PlansPageBase: React.FunctionComponent<IPlansPageBaseProps> = ({
                 {!planList ? null : planList.length === 0 ? (
                   <EmptyState variant="full">
                     <EmptyStateIcon icon={AddCircleOIcon} />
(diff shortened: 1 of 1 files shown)


vconzola:
@mturley Where is this page coming from? The migration plans empty state should direct users to the Clusters page if fewer than two clusters have been added. See slide 16 of the mockups. https://docs.google.com/presentation/d/1JU_Lw1A-gUO4gBRofZzi74aONcknYRHbqBl_xHxikYQ/edit?usp=sharing

eriknelson:
@vconzola this is all still heavily under construction and left over from the previous UI without the left nav. I would not get too concerned if you see things like this right now.

mturley:
@vconzola sorry, I hadn't actually realized that.. I haven't gone back over the mockups carefully yet, I am planning to go back and make sure I nail down all of these details after we get the main flow working.

PR #932, opened 2020-06-18
Add admin / non-admin role indicator in top-right of masthead

mturley:
Screenshot 2020-06-18 18 16 34

Screenshot 2020-06-18 18 17 06


vconzola:
@mturley @eriknelson Is this a permanent addition or just there during development?

mturley:
Oh, most definitely just during development. We wanted to make it clear that the stuff we demo tomorrow is incomplete.

mturley:
Oh, sorry, wrong PR 😄 I think it's probably temporary, until we have a proper identity/role dropdown there, is that right @eriknelson @ibolton336 ?

PR #933, opened 2020-06-19
Typescript bump

gildub:
Typescript is upgraded to  3.9.5

TS was due for an upgrade but also there are deps needs for testing framework jest and @testing-library which requires Typescript > 3.8.3
Using new version exposes a couple of issues which have been addressed:
* TS2305 with app/home/duck/index
* TS2339 with app/plan/duck/selectors.ts

And also following `app/home/duck/index` fix makes use of the barrel file (index) for `{context|hooks}` imports.

### DIFF
--- a/src/app/common/components/ErrorModal.tsx
+++ b/src/app/common/components/ErrorModal.tsx
@@ -12,7 +12,7 @@ import { ErrorCircleOIcon } from '@patternfly/react-icons';
 import { useHistory } from 'react-router-dom';
 import { AlertActions } from '../duck/actions';
 import { connect } from 'react-redux';
-import { PollingContext } from '../../home/duck/context';
+import { PollingContext } from '../../home/duck';
 const styles = require('./ErrorModal.module');
 
 interface IProps {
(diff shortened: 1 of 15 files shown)


ibolton336:
any reason to not try to bump to latest if we are bumping? Looks like 3.9.5 is the latest version

gildub:
@ibolton336, I was trying to push it up without being too disruptive but that's a good point, I haven't tried. Let's see...

src/app/common/components/ErrorModal.tsx
```diff
@@ -12,7 +12,7 @@ import { ErrorCircleOIcon } from '@patternfly/react-icons';
 import { useHistory } from 'react-router-dom';
 import { AlertActions } from '../duck/actions';
 import { connect } from 'react-redux';
-import { PollingContext } from '../../home/duck/context';
+import { PollingContext } from '../../home/duck';
```
ibolton336, line 15: This is throwing an error for me. Any reason we are changing the path here?

gildub: The reason is because before we're defining a barrel file (index file loading deps) but we're not using it. If we use `.../home/duck/context` then there is no need for the index file.

gildub:
Restart your IDE to refresh paths.

ibolton336:
> Restart your IDE to refresh paths.

:+1:

PR #938, opened 2020-06-24
Test for GeneralForm Component

gildub:


mturley:
@gildub is this `do-not-merge` temporary / do you plan to continue with this PR later on? If not let's close it

gildub:
@mturley,

This how it looks like after the much easier approach with  `useFormikContext`.
Testing each step of the wizard allows to test adding or updating input fields only since the validation button are wrapped at the wizard level.
We could also add field validation.
What do you think?

mturley:
@gildub, this looks great! I do think field validation might be a good thing to test, even though we can't check the button validation at this level. That does make me think we should have one big integration test that mounts the wizard and steps through all the steps to make sure the button is enabled/disabled properly. Although for that, we might be able to still have a test for each step, and mount `<WizardFormik><WizardComponent .../></WizardFormik>` in the test so we can populate it with initial values and just move to the step in question and check the button. Up to you, and that can be a different PR.

For this one, there are only a couple small things I would change, to make the test file a little more lightweight:
* You can import `defaultInitialValues` from `WizardContainer` to use for your `initialValues` prop, that way if we change the real initial values your tests will still be accurate. If you want to override a value you can still do `{ ...defaultInitialValues, someField: 'someValue' }`.
* We should probably make a separate file for mocks/fixtures such as your `clusterList` and `storageList` here, and import those in so we can use them in multiple tests. We could put them some place common so they can also be used to populate a mock redux store for bigger integration tests and such.

gildub:
@gildub, thanks for the initial review. I've followed your suggestions.

mturley:
Should we remove the `do-not-merge` label on this one?

PR #939, opened 2020-06-24
Fix isLoadingNewToken

gildub:
Found out the following message while running tests:
`Warning: Failed prop type: Invalid prop `isDisabled` of type `string` supplied to `SelectToggle`, expected `boolean`.`

Changing IsLoadingNewToken to be boolean fixes it.


gildub:
This is obsolete

PR #940, opened 2020-06-24
Fix bugs with token creation flow in plan wizard, form validation on general step, and polling in token modal

### LINKED ISSUE #930: Bug in non-admin plan wizard general step: adding a new token does not properly select it
The logic for determining if we are waiting for the newly created token to load is flawed, adding a token gets you stuck in that state. Also, if the token you added is in a different cluster than the selected one, it never appears in the dropdown so we need to handle that case. We may need separate state to store the name of the just-created token, and only put it in the form value if we get back one in the tokenList that matches.

Also, the token modal stops all polling when it opens, and does not resume it when it closes. This is fine in the wizard because we don't want polling in the wizard, but it's a problem on the tokens page. We may need a `stopAndResumePolling` prop or whatever in the AddEditTokenModal so that behavior can be different based on where it's rendered.

mturley:
Fixes #930 

When the user is creating a plan and there are no available tokens for the selected cluster, they can open the add token modal right from the token select dropdown and then have that token selected once it is finished being added/loaded. My previous logic for this was flawed: the token name from the modal was immediately set in Formik state on submitting the modal, but the token doesn't immediately load into `tokenList` in redux so we'd end up in a state where the selected value is not one of the available options. Also, if somehow the created token does not match the selected cluster, we'd be stuck with that "loading" message indefinitely.

This PR fixes that issue by keeping the name of the token being created in state and waiting for a matching token to load before trying to set the Formik value. If the new token loads but does not match the selected cluster, it gives up and just hides the loading message. I also made that message a little nicer-looking by rendering it inside the placeholder text of the Select component, with a spinner and "Adding token..." instead of "Loading...":

Screenshot 2020-06-24 15 03 53

See the fixed flow in action:

q1u3UXQgwj

This PR also fixes the behavior of the AddEditTokenModal around polling. Previously when it was opened it would stop polling, but never resume it on close. We want polling to be paused while the modal is open and resumed on close, but when it is open inside the wizard we don't want to resume polling on close because the wizard also keeps polling paused. (Modal-in-a-modal is a fun use case 😄 ) So, this PR adds a prop `preventPollingWhileOpen` that defaults to true, and has a false value when mounted in the wizard. As part of this, I just tied that polling pause/resume to the component being mounted and unmounted, so instead of mounting it all the time and using the `isOpen` prop to hide/show the modal, I only mount the modal when `isOpen` would be true. (this also results in less cruft in the `<body>` caused by [the z-index hack](https://github.com/konveyor/mig-ui/pull/914/files#diff-30cc324d6d215c2355f893820ccd6dc8R37); that wrapper will only be present when the modal is open instead of all the time).

Lastly, there was an issue where selecting a source or target cluster would cause the "Required" validation message to appear even though there is a selected value, and it would only go away when some other field was changed. It seems like this was caused by the `setFieldTouched` method being called on change instead of on blur, so I fixed that. I also cleaned up a few things around TypeScript types for `SimpleSelect`, to enable non-string values in the `OptionWithValue` type (yay, [generics](https://www.typescriptlang.org/docs/handbook/generics.html)!)

### DIFF
--- a/src/app/common/components/AddEditTokenModal/AddEditTokenForm.tsx
+++ b/src/app/common/components/AddEditTokenModal/AddEditTokenForm.tsx
@@ -144,10 +144,7 @@ const InnerAddEditTokenForm: React.FunctionComponent<
         />
         <div className={`${spacing.mtMd} ${spacing.mbLg} ${spacing.mlLg}`}>
           <TextContent>
-            <Text component="p">
-              Log in to the cluster to automatically generate the token, or enter the token string
-              in the field.
-            </Text>
+            <Text component="p">Log in to the cluster to automatically generate the token.</Text>
           </TextContent>
           <Button
             variant="secondary"
--- a/src/app/common/components/AddEditTokenModal/AddEditTokenModal.tsx
+++ b/src/app/common/components/AddEditTokenModal/AddEditTokenModal.tsx
@@ -18,7 +18,7 @@ interface IAddEditTokenModalProps {
   clusterList: ICluster[];
   addToken: (tokenValues: ITokenFormValues) => void;
   onClose: () => void;
-  onTokenCreated?: (tokenRef: INameNamespaceRef) => void;
+  onTokenAdded?: (tokenRef: INameNamespaceRef) => void;
   preSelectedClusterName?: string;
   migMeta: IMigMeta;
 }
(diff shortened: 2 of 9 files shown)


mturley:
This will invalidate https://github.com/konveyor/mig-ui/pull/939, so if we merge it before that we can close #939. If it needs time for review though, #939 is worth merging to remove that warning in the meantime, and I'll rebase.

ibolton336:
Seeing this break the next button validation from the general step in the wizard. The Next button is disabled with no apparent validation errors?

ibolton336:
issue1

mturley:
Crap, I think I know what I did there. Sorry, I didn't have a token for my target cluster so I faked that data to test it, I guess I didn't fake it well 😆

mturley:
Opened https://github.com/konveyor/mig-ui/issues/941 to track the issues we started seeing here

PR #944, opened 2020-06-26
Add test helper utils to wrap redux store

gildub:
Utils helper to be used testing UI components.

Follows https://redux.js.org/recipes/writing-tests#connected-components


ibolton336:
> Maybe it makes more sense to just have a mock store in all of our tests anyway so we don't have to think about what does and does not use redux. What do you think @ibolton336 ?

I think if its simple enough to abstract out 2 functions that would definitely add more clarity & would be a nice touch.

gildub:
+1 using a different name such as `renderWithStore` to make clear distinction.

I don't think using a mock store is going to work because every component we're going to test is going to be composing component requiring being wrapped by a provider .
And we're not going to test leaf component in respect of the philosophy of the testing library to test UI from an end user perspective.

src/test/utils.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { render as rtlRender } from '@testing-library/react';
+import { createStore } from 'redux';
+import { Provider } from 'react-redux';
+import rootReducer from '../reducers';
+
+function render(
```
mturley, line 7: Are we still planning to rename this as `renderWithStore` or something similar?

mturley: @gildub if you rename this function I'll go ahead and merge this, assuming you still plan to use it. Otherwise let's close it

ibolton336:
Closing due to lack of activity.

PR #949, opened 2020-06-29
AddEditClusterModal initial component test

gildub:
Contains 3 types of test examples:
- Allows filing a cluster with valid values: The "Add cluster" button should be enabled
- Triggers validation messages when using invalid values: The "Add cluster" button should be disabled
- Loads a form with provided values and allows changing form: The "Update cluster" button should be enabled

### DIFF
--- a/src/app/home/pages/ClustersPage/components/AddEditClusterModal/AddEditClusterForm.tsx
+++ b/src/app/home/pages/ClustersPage/components/AddEditClusterModal/AddEditClusterForm.tsx
@@ -102,6 +102,7 @@ const InnerAddEditClusterForm = (props: IOtherProps & FormikProps<IFormValues>) 
         {/*
           // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
         <TextInput
+          data-testid="cluster-name"
           onChange={formikHandleChange}
           onInput={formikSetFieldTouched(nameKey)}
           onBlur={handleBlur}
@@ -157,6 +158,7 @@ const InnerAddEditClusterForm = (props: IOtherProps & FormikProps<IFormValues>) 
         {/*
           // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
         <TextInput
+          data-testid="url"
           onChange={formikHandleChange}
           onInput={formikSetFieldTouched(urlKey)}
           onBlur={handleBlur}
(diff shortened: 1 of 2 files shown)


src/app/home/pages/ClustersPage/components/AddEditClusterModal/__tests__/AddEditClusterModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { createStore } from 'redux';
+import { Provider } from 'react-redux';
+import rootReducer from '../../../../../../../reducers';
+
+import AddEditClusterModal from '../index';
+
+const store = createStore(rootReducer, {});
+
+describe('<AddEditClusterModal />', () => {
+  it('adds a cluster with valid values', () => {
```
mturley, line 15: Small nit-pick, since this is just testing form validation and not actually adding a cluster, maybe we should call it something like "enables add cluster button when values are valid"?

src/app/home/pages/ClustersPage/components/AddEditClusterModal/__tests__/AddEditClusterModal.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <AddEditClusterModal {...initialProps} />
+      </Provider>
+    );
+
+    userEvent.type(screen.getByTestId('url'), 'example.com');
+
+    expect(screen.getByRole('button', { name: 'Add cluster' })).toBeDisabled;
+  });
+
+  it('updates a cluster with valid fields', () => {
```
mturley, line 69: same here, maybe "properly fills initialClusterValues and enables add cluster button when valid" or something. Not necessarily that, just thinking ahead in case we want to add more integration-style tests that involve redux actions to actually add/update the cluster.

src/app/home/pages/ClustersPage/components/AddEditClusterModal/AddEditClusterForm.tsx
```diff
@@ -181,12 +182,13 @@ const InnerAddEditClusterForm = (props: IOtherProps & FormikProps<IFormValues>)
         {/*
           // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
         <TextInput
+          data-testid="account-token"
```
mturley, line 185: Sorry I missed this before, but since we're not using this data-testid anymore can you remove it?

PR #952, opened 2020-07-01
Non-admin changes for Clusters page (empty state, add token action, blocking add/edit/remove)

mturley:
Closes https://github.com/konveyor/mig-ui/issues/892.

* Adds an alternate empty state for non-admins, indicating that an admin needs to add their clusters for them.
* Adds an action "Add token" to the dropdown menu on each cluster's row in the ClustersTable which opens the AddEditTokenModal with the corresponding cluster pre-selected
* Prevents non-admin users from adding, editing or removing clusters (hides Add button and Edit/Remove actions, prevents AddEditClusterModal from rendering at all)

Screenshot 2020-06-30 16 59 49

Screenshot 2020-07-06 16 54 24

Screenshot 2020-07-06 16 54 38


mturley:
I need to change one quick thing here, don't merge just yet

mturley:
Ok, ready to go again.

PR #954, opened 2020-07-02
Component testing: Add filling forms for AWS S3, GCP and Azure

gildub:
- Replaces `getByTestId` with `getByLabelText` per react testing library (dom) recommendation by using `aria-label` instead of `data-testid` tag. This also has the advantage of not using a tag only dedicated for tests.
- Replaces (and uses) `getByRole` with `getByTestId` for the primary submit button. This allows to easily identify the button for both "Add" and "Update" repository cases.
- Adds test cases for AWS S3, GCP and Azure storages.
- Fix `input` fields `id` to match `label` `for` (or `fieldId`)

### DIFF
--- a/src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/GCPForm.tsx
+++ b/src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/GCPForm.tsx
@@ -179,6 +179,7 @@ const InnerGCPForm = (props: IOtherProps & FormikProps<IFormValues>) => {
           {addEditButtonTextFn(currentStatus)}
         </Button>
         <Button
+          aria-label="GCP Storage Submit Form"
           variant="secondary"
           isDisabled={isCheckConnectionButtonDisabled(
             currentStatus,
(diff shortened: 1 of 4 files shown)


src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/GCPForm.tsx
```diff
@@ -128,11 +129,13 @@ const InnerGCPForm = (props: IOtherProps & FormikProps<IFormValues>) => {
           // @ts-ignore issue: https://github.com/konveyor/mig-ui/issues/747 */}
 
         <TextInput
+          aria-label="GCP bucket name"
```
mturley, line 132: Interesting, this shouldn't be necessary if the `label` and `fieldId` props of `FormGroup` correspond to this field properly. Looks like we screwed that up though, the `id` on `TextInput` should match the `fieldId` on `FormGroup`, and that way the label in the `FormGroup` should apply to the input in the HTML.

Maybe try changing this field's `id` to `{gcpBucketKey}` to match the `fieldId`, and remove the `aria-label` prop and see if your test still selects it properly? It would be nice to not have to duplicate these label strings. If that doesn't work I'm fine with this.

gildub: @mturley, thanks for this, that's crystal clear now. I've finally put the whole thing together. I was missing a regex to select label or roles properly:
https://spectrum.chat/testing-library/help/updated-to-v3-0-2-and-getbylabeltext-isnt-working~70994f71-252c-4437-bcbd-49425298952a

mturley: This is awesome! I'm really digging the approach of selecting fields based on accessible labels. In fixing this you're resolving an accessibility issue for screen-reader users that we never would have otherwise found.

PR #956, opened 2020-07-02
Fix regression with value check on token option select

ibolton336:


mturley:
Thanks @ibolton336

mturley:
Looks like this Travis issue might be related to Travis running into GitHub's API rate limiting when using their legacy commit status API. Looks like that's what we're using ("continuous-integration/travis-ci" rather than the new GitHub App for Travis which calls the check "Travis CI - Pull Request") and I found some related issues:
https://travis-ci.community/t/fails-to-report-build-status-to-github-again/7880
https://travis-ci.community/t/github-status-not-posted-on-commits-on-repositories-using-legacy-service-integration/7798/14
Sounds like the recommended solution is to migrate to their newer API (move the project from travis-ci.org to travis-ci.com): https://travis-ci.com/account/migrate

mturley:
It looks like the mig-controller repo has already been migrated to the new Travis integration, too.

PR #967, opened 2020-07-15
Move login OAuth token handling to backend

mturley:
~WIP. Has issues with self-signed cert error handling.~

Cert issues resolved & ready for review. This PR moves the oauth flow from the react loginComponent to a custom node server. The node server will intercept proxied routes to /login & /login/callback & push the user through an authorization code grant oauth2 flow.


eriknelson (approved):
Have this tested and working in my environment after a bunch of back and forth. I think this is ready to go!

PR #968, opened 2020-07-15
Add edit token functionality

### LINKED ISSUE #973: Bug: fix associated cluster for  add/edit token flow.
Currently the associated cluster is not being updated properly when adding a token. Need to make sure the value is correct.

### LINKED ISSUE #974: Update the token table status with real-time status from api
Need to add icons & warning messages for token table. Currently all tokens show an OK status message with a green check.

ibolton336:
Adds the ability to edit a MigToken from the tokens table. Currently, the only editable field is SA token until oauth is implemented. When you open the edit modal, all other fields are disabled including the ability to change the token type. 

Closes #974 
Closes #973

### DIFF
--- a/src/TEMPORARY_GLOBAL_FLAGS.ts
+++ b/src/TEMPORARY_GLOBAL_FLAGS.ts
@@ -1,4 +1,4 @@
 // NATODO delete this file and remove all references when we are ready to release non-admin support
 // (will need to manually revert changes from https://github.com/konveyor/mig-ui/pull/966)
 
-export const NON_ADMIN_ENABLED = false;
+export const NON_ADMIN_ENABLED = true;
--- a/src/TEMPORARY_GLOBAL_FLAGS.ts
+++ b/src/TEMPORARY_GLOBAL_FLAGS.ts
@@ -1,4 +1,4 @@
 // NATODO delete this file and remove all references when we are ready to release non-admin support
 // (will need to manually revert changes from https://github.com/konveyor/mig-ui/pull/966)
 
-export const NON_ADMIN_ENABLED = true;
+export const NON_ADMIN_ENABLED = false;
--- a/src/app/home/pages/ClustersPage/components/ClusterActionsDropdown.tsx
+++ b/src/app/home/pages/ClustersPage/components/ClusterActionsDropdown.tsx
@@ -14,6 +14,8 @@ interface IClusterActionsDropdownProps {
   clusterInfo: IClusterInfo;
   removeCluster: (clusterName: string) => void;
   isAdmin: boolean;
+  toggleAddEditTokenModal: () => void;
+  isAddEditTokenModalOpen: boolean;
 }
 
 const ClusterActionsDropdown: React.FunctionComponent<IClusterActionsDropdownProps> = ({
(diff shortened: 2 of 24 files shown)


src/TEMPORARY_GLOBAL_FLAGS.ts
```diff
@@ -1,4 +1,4 @@
 // NATODO delete this file and remove all references when we are ready to release non-admin support
 // (will need to manually revert changes from https://github.com/konveyor/mig-ui/pull/966)
 
-export const NON_ADMIN_ENABLED = false;
+export const NON_ADMIN_ENABLED = true;
```
mturley, line 4: We should leave this set to false and just flip it to true as an uncommitted change while testing this stuff, so we don't accidentally flip it to true on master

ibolton336: Ah! yeah i keep accidentally committing this.

ibolton336: Ended up just locally ignoring that file with: 

git update-index --skip-worktree src/TEMPORARY_GLOBAL_FLAGS.ts

ibolton336:
TODO: 

#974 
#973 
#972

mturley:
@ibolton336 are you thinking we should merge this as-is or are you planning to address those TODO issues in the same PR?

src/TEMPORARY_GLOBAL_FLAGS.ts
mturley: Ah! Didn't know you could do that, nice

src/client/resources/conversions.ts
```diff
@@ -43,14 +43,22 @@ export function getTokenSecretLabelSelector(
...
-export function updateTokenSecret(rawToken: string) {
+export function updateTokenSecret(rawToken: string, isMigTokenSecret?: boolean) {
   // btoa => to base64, atob => from base64
   const encodedToken = btoa(rawToken);
-  return {
-    data: {
-      saToken: encodedToken,
-    },
-  };
+  if (isMigTokenSecret) {
```
mturley, line 49: What's this `isMigTokenSecret` flag for? I noticed it's hard-coded to `true` above for now

src/app/home/pages/ClustersPage/components/ClusterActionsDropdown.tsx
```diff
@@ -69,6 +74,7 @@ const ClusterActionsDropdown: React.FunctionComponent<IClusterActionsDropdownPro
                   onClick={() => {
                     setKebabIsOpen(false);
                     toggleAddEditTokenModal();
+                    setAssociatedCluster(clusterName);
```
mturley, line 77: What's the reason for moving `associatedCluster` / `setAssociatedCluster` into redux and then using that in the modal instead of having the `preSelectedClusterName` prop passed to the `<AddEditTokenModal />` below?

mturley:
Aside from my couple of questions above, this code looks good to me! I'm having a bit of trouble getting my environment to play nice with your cluster though @ibolton336 so I can't test it... I sent you details on slack :(

PR #976, opened 2020-07-22
AddEditClusterFormModal props and interface refactor

gildub:
- Defines interface for props
- Removes unused cluster props

### DIFF
--- a/src/app/home/pages/ClustersPage/components/AddEditClusterModal/index.tsx
+++ b/src/app/home/pages/ClustersPage/components/AddEditClusterModal/index.tsx
@@ -10,6 +10,7 @@ import {
   AddEditState,
 } from '../../../../../common/add_edit_state';
 import { PollingContext } from '../../../../duck/context';
+import { IReduxState } from '../../../../../../reducers';
 
 // TODO add types here
 const AddEditClusterModal = ({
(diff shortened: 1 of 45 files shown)


src/app/home/pages/ClustersPage/components/AddEditClusterModal/index.tsx
```diff
@@ -1,54 +1,62 @@
...
+  addCluster,
+  cancelAddEditWatch,
+  onHandleClose,
+  resetAddEditState,
+  updateCluster,
+}: IAddEditClusterModal) => {
   if (!isAdmin) return null;
 
   const pollingContext = useContext(PollingContext);
-  const [currentClusterName, setCurrentClusterName] = useState(
```
ibolton336, line 30: Removing this breaks the add edit cluster flow. There is an opportunity to refactor this but removing it leaves the add/edit/check connection in a broken state.

ibolton336: Hey Gilles - submitted a PR to address some changes i ran into during my add/edit token work. Lines up nicely with what you have already started here. See PR https://github.com/gildub/mig-ui/pull/6

PR #978, opened 2020-07-28
Refactor plan status

ibolton336:
Working toward resolving https://bugzilla.redhat.com/show_bug.cgi?id=1837656

### DIFF
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -1,6 +1,6 @@
 import { createSelector } from 'reselect';
 import moment from 'moment-timezone';
-import { filterPlanConditions, filterMigrationConditions } from './helpers';
+import { filterPlanConditions, filterLatestMigrationConditions } from './helpers';
 
 const planSelector = (state) => state.plan.migPlanList.map((p) => p);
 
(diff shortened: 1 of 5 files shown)


ibolton336:
image

src/app/plan/duck/selectors.ts
```diff
@@ -102,134 +103,54 @@ const getPlansWithPlanStatus = createSelector(
...
-              m.status.conditions.some((c) => c.type === 'Succeeded') &&
-              !latest.status.conditions.some((c) => c.type === 'Canceled')
-            );
-          }
-        }).length;
+      const hasSucceededMigration = !!plan.Migrations.filter((m) => {
+        if (m.status && !m.spec.stage) {
+          return (
+            m.status.conditions.some((c) => c.type === 'Succeeded') &&
+            !latestMigration.status.conditions.some((c) => c.type === 'Canceled')
```
mturley, line 125: I know this was already present in the existing logic, but I just noticed that `hasSucceededMigration` is checking for a canceled condition on only the latest migration here every time, but above in `hasSucceededStage` it checks for a canceled condition on the current migration being checked by filter predicate. Is that intentional? Not sure I fully grok the logic.

ibolton336: Good question - the issue is that on Cancelled migrations/stages, there is sometimes a succeeded condition also.

ibolton336:
> @ibolton336 the code looks good to me and passes basic sanity checks, but I can't thoroughly test it since I don't have real migrations in my environment of the various statuses. If you've tested that those all work right I'm fine with merging, or if you can help me figure out how to test them that works too

Good question with

> @ibolton336 the code looks good to me and passes basic sanity checks, but I can't thoroughly test it since I don't have real migrations in my environment of the various statuses. If you've tested that those all work right I'm fine with merging, or if you can help me figure out how to test them that works too

Part of the reason why this is a tough area of the code to touch is because of all of the different types of migrations we can do. 3.7-> 4.x, 4.x -> 3.11, 4.x -> 4.x, etc. It takes a great deal of time to test all of these scenarios.

ibolton336:
> @ibolton336 Where did that icon come from for the last state = failed? It should be fa-exclamation-circle.
> 
> And I know I've been saying this for months, but we as a team, really need to sit down and go through what we display for status/state throughout the the UI.

image
How is does this look? Agreed we should refine this further when time permits.

vconzola:
@ibolton336 I like that better. :-) Thanks.

ibolton336:
Screen Shot 2020-07-29 at 5 21 26 PM

PR #981, opened 2020-07-30
fix secret reference for update cluster request

ibolton336:
This is a small regression from the latest changes to the add/edit cluster flow. Currently a cluster update fails due to the wrong reference to an existing secret on a cluster. This PR will fix this bug.


ibolton336:
Need to verify if this is occurring on the latest mig-controller image. Currently using a preliminary image for namespace analytics which may be the cause.

ibolton336:
Confirmed the issue exists on latest images. Ready to merge.

PR #985, opened 2020-08-06
Add TreeView and DebugResourcesModal components with mock data

mturley:
Based on a discussion with @eriknelson and @vconzola, we needed a way to present a migration plan's associated resources as a tree view for debug purposes (detailed here: https://github.com/konveyor/enhancements/pull/2/files)

Initially I started by using an off-the-shelf tree component since the PatternFly TreeView component is not yet available, but I realized that I could implement a true PatternFly tree view after all by copying [the unreleased CSS](https://github.com/patternfly/patternfly/pull/3354). We don't need the advanced features (search, checkboxes, icons, badges, actions), so the HTML structure itself is not very complex and it makes for a fairly small component. I made a few tweaks to the copied CSS so that we can use it with our current version of PatternFly.

I intend to eventually add those missing features to this TreeView component and contribute it upstream to PatternFly, assuming someone else doesn't beat me to the React implementation. Either way, we'll be able to upgrade PatternFly and swap in that official component once it's available, without changing the look and feel.

This implementation uses mock data, we'll need to swap in real data from Redux when that's available. Perhaps we should comment out this new action button and merge this PR, then reveal it in another PR when the real data is available.

# Screens

The new view is in a modal accessible from the kebab menu on a migration plan:
Screenshot 2020-08-05 18 02 36

And here it is:
Screenshot 2020-08-05 20 27 07

### DIFF
--- a/src/app/common/components/TreeView/TreeView.module.scss
+++ b/src/app/common/components/TreeView/TreeView.module.scss
@@ -1,0 +1,194 @@
+// This is based on the unreleased PatternFly tree view CSS found at https://github.com/patternfly/patternfly/blob/master/src/patternfly/components/TreeView/tree-view.scss.
+// We should remove it after we upgrade our PatternFly dependencies and start using the real component.
+
+@mixin pf-text-overflow {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+$pf-c-tree-view--MaxNesting: 10;
+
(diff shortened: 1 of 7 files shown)


mturley:
For reference, here is the new PatternFly issue for the upstream React implementation: https://github.com/patternfly/patternfly-react/issues/4652

eriknelson:
Mike, thanks very much for this. I'll be picking it up today.

PR #986, opened 2020-08-07
Namespace resources/ migAnalytics support

ibolton336:
Includes support for the new migAnalytics feature. This allows a user to view a detailed list of namespaces within the mig plan table. 

- Support for editing migPlan namespaces list
- Ability to refresh migAnalytics manually with a button click
- Auto cleanup of migAnalytics objects on plan delete

mig-analytics-progress


ibolton336:
Blocked by: https://github.com/konveyor/mig-controller/pull/604

vconzola:
@jmontleon Looking at the screen recording above, don't know if this is real data be retrieved, but it seems to be taking a long time for a single namespace without much in it. Is it possible to provide better progress on the collection process?

ibolton336:
loading-updates
@vconzola

ibolton336:
Peek 2020-08-20 13-57

ibolton336:
add-ns-icon

vconzola:
@ibolton336 Thanks. One thing... It's hard to tell for sure from the recording, but it looks like the row is being expanded immediately when the plan is added to the list? It shouldn't be. We want the user to have to click to expand the row. The analytics should run in the background immediately, but we don't want the user to see the data until they click to expand the row. All part of minimizing the time a user has to wait for the namespace data to be retrieved.

ibolton336:
not-opening-by-default
@vconzola It is not expanding immediately. The expanded state was still open before opening the wizard to edit. Sorry for the confusion.

vconzola:
@ibolton336 Cool. Thanks.

src/app/home/pages/PlansPage/components/AnalyticsTable.tsx
```diff
@@ -0,0 +1,256 @@
...
+  useEffect(() => {
+    const namespaces = latestAnalytic?.status?.analytics?.namespaces
+      ? latestAnalytic?.status?.analytics?.namespaces
+      : [];
+
+    setIsLoadingAnalytics(
+      (analyticPercentComplete !== 100 && latestAnalytic) || isRefreshingAnalytic ? true : false
+    );
+
+    const mappedRows = namespaces.map((namespace) => {
```
eriknelson, line 78: Looks like this could blow up on a map call because `namespaces` could be null from a short-circuit return on 70. Do we need to default mapped rows to an empty state if namespaces is null?

It actually feels like with the introduction of this syntax it's really easy to ignore what happens if that returns a null and creates a lot of opportunity for calls on a null value, if I'm understanding that correct?

src/app/home/pages/PlansPage/components/PlansTable.tsx
```diff
@@ -73,15 +99,24 @@ const PlansTable: React.FunctionComponent<IPlansTableProps> = ({
...
         migrationCount,
         sourceClusterName,
         targetClusterName,
         storageName,
-        pvCount,
+        namespaceCount,
+        isMaxResourcesLimitReached,
       } = getPlanInfo(plan);
+      const isLoadingAnalytic: boolean =
+        !!(plan?.PlanStatus?.analyticPercentComplete !== 100 && plan.PlanStatus.latestAnalytic) ||
```
eriknelson, line 113: I feel like we need to unwrap this logic to be more explicit, this is pretty gnarly to try to reason through just reading it. Like, just trying to figure out what `isLoadingAnalytic` from this whole boolean expression is not clear.

eriknelson (changes requested):

src/app/home/pages/PlansPage/components/AnalyticsTable.tsx
ibolton336: @eriknelson If you look above on line 72, 


```
 const namespaces = latestAnalytic?.status?.analytics?.namespaces
      ? latestAnalytic?.status?.analytics?.namespaces
      : [];
```

I think that should take care of any null values.

src/app/home/pages/PlansPage/components/PlansTable.tsx
ibolton336: You're right - I will refactor this.

src/app/home/pages/PlansPage/components/AnalyticsTable.tsx
eriknelson: Yep, you're right. Mssed that.

eriknelson (approved):

PR #987, opened 2020-08-07
Update header for rebrand

ibolton336:
image

Addresses https://issues.redhat.com/browse/MIG-294


mturley:
Also @ibolton336 that `` should have `vertical-align: middle;` applied to it as well, without that it sits a couple of pixels higher than the middle of the masthead.

PR #990, opened 2020-08-19
Upgrade patternfly - Major version bump to 4.x.x

ibolton336:
Upgrading our PF4 packages will allow us greater flexibility when creating & sharing modules across MTV & MTC. 

This PR addresses all breaking changes stemming from pf api changes when bumping to PF4.


ibolton336:
image
image

ibolton336:
Screen Shot 2020-08-26 at 11 52 14 AM

src/app/home/pages/PlansPage/components/Wizard/WizardComponent.tsx
```diff
@@ -276,8 +276,6 @@ const WizardComponent = (props: IOtherProps) => {
           title="Create a migration plan"
           onClose={handleClose}
           steps={steps}
-          isFullWidth
```
ibolton336, line 279: @mturley Any alternative for this prop? Looks like it went away with pf 4.x

mturley: It looks like it was just removed without an alternative, I assume because it's now the default behavior? I'll look into it.

mturley: @ibolton336 I figured it out. Wizard was modified so that it can be used directly in a page layout or wrapped in a Modal. It still automatically renders in a modal when `isOpen` is passed for those who just need the default Modal props, but you can pull it out and wrap it in your own modal and pass a `width` to that modal. Here's a fix: https://github.com/mturley/mig-ui/commit/e386a3b9204a3b8d3978c2d50a448a393e3957c6

ibolton336: Thanks mike - looks much better 👍

mturley:
@eriknelson if you're cool with this as-is I think I'll merge it

PR #994, opened 2020-08-23
Integrate the DebugTreeView with discovery service

eriknelson:
This PR is based on top of #985. It integrates the UI's DebugTreeView that Mike brought in from PF with the new discovery service debug tree, as outlined in https://github.com/konveyor/enhancements/tree/master/enhancements/debug#proposed-ui-debug-experience. It can't be tested without @jortel's custom discovery service image which provides the backend: `quay.io/jortel/mig-controller:tree`.

I would consider this PR still a WIP -- there's a decent amount of UX polish that's required, but the foundation is there to show what's possible, and it's useful in its current form. Namely:

* It displays the tree of all relevant objects for a given plan and its migrations.
* You can use the one-click copy helper to copy the relevant `oc get` command to your clipboard for any of the objects.
* It has a raw JSON view for any of the objects that's easily accessible. Right now, this is literally a new blank page with an enhanced JSON document viewer, but we can relatively easily modify that depending on desired UX.

## Issues

Let's discuss what's in-scope for this PR and should be fixed before accepting a merge, and what can be filed as an issue:

* Styles are jacked up. Need some help with that.
* @mturley I'm certain I've put a bunch of stuff in the TreeView components that doesn't belong there, particularly if we want that generic and submitted upstream. Let's work together to get stuff organized in the right place.
* I'm not convinced we actually want to do this, but it'd be possible to poll instead of a manual refresh.
* This is basically a UX first draft @vconzola, I think it would be best for you to file an issue with everything that you would like to see changed about this prior to shipping for 1.3, and we'll make sure that happens. I'm thinking that might be significant enough that it makes sense to do in a follow up PR :)
* **Important:** Viewer will error out with network error if you try to view the tree prior to running through the migplan wizard (which is the way that you indirectly accept the discovery service network cert). We discussed in the past that the discovery service cert issue would be solved naturally due to non-admin flow, but now that non-admin has taken a back seat, we need to come up with a plan so that people can accept this cert early in the flow. A dummy request on app startup may be the easiest solution and best user experience since it will be just another cert acceptance in the line of them.
* Would be nice to have a live status in the modal so you can see where the migration is at while the modal is open?
* When you're in the raw view, naturally you will go back in the browser and want to be where you left off. Instead, you're dropped back to the plans screen and need to open the window again.

## Screenshots

**Prior to migrations**
PriorToMigrations

**Raw plan view**
RawPlanView

**Migration Started**
MigrationStarted

**Initial Backup Started**
InitialBackupStarted

**Second Backup Started**
2ndBackupStarted

**Restore Begun**
RestoreBegun

**Final Tree**
FinalTree

**Copy `oc get` command**
CopyOCGet

**Resulting `oc get` command**
OcGetCmd


mturley (approved):
This looks good to me as-is for now, we can follow up on the UX changes in ﻿https://github.com/konveyor/mig-ui/issues/999 using the new upstream https://www.patternfly.org/v4/documentation/react/components/treeview .

PR #1007, opened 2020-09-09
Improve plan debug resources tree view UX

### LINKED ISSUE #999: UX changes for DebugTreeView
After reviewing the current implementation in PR #994 we suggest the following UX changes:

- When user selects the "debug" action for a plan open the DebugTreeView in a new page - ideally in a new tab, so the user can switch back and forth between the migration UI and the debug view.
- Add a toolbar to the tree view with search capability and a Refresh button.
- Fix the alignment of the tree rows with indentation so that the hierarchy is more obvious.
- Remove the gray background from the tree rows.
- Put the actions for each object (tree row) in a kebab. Two actions are available:
   1. - Get CLI command (Tooltip??: Generate "oc get" command for the object and copy it to the clipboard)
   2. - View JSON (Tooltip??: View the raw JSON for the object)  Opens the JSON viewer in a modal. (Not sure if this is better than a new page or possibly even a drawer. Need to discuss.)

We should be able to get a lot of these changes for "free" by pulling in the new PF tree view component.

mturley:
Closes #999.

Screen Shot 2020-09-09 at 7 19 30 PM

Screen Shot 2020-09-09 at 7 20 19 PM

* Moves the debug view from its modal onto a new page routed at `/plans/:planName/debug`, which opens in a new tab
* Adds a search bar and refresh button
* Removes the TreeView clone we built here and replaces it with the new PatternFly TreeView (upgrades pf packages in the process). This fixes the gray background and indentation issues.
* Move the CLI command copy and view JSON actions into a kebab menu.

@vconzola, I didn't move the JSON view into a modal, it still opens in a new page since that was already implemented in the router. If we determine that isn't ideal we can move it to a modal, but it sounded like that decision wasn't final.

Note: due to a restriction with the `action` property on TreeView items expecting you to pass children for its internal `<button>` element, this currently renders the kebab as a button-in-a-button. We should fix this, but it will require a change in PatternFly (I opened https://github.com/patternfly/patternfly-react/issues/4799). In the meantime, there is a console warning complaining about this invalid button nesting structure, but it won't cause any problems in production.


mturley:
Opened https://github.com/konveyor/mig-ui/issues/1008 to follow up about the HTML structure warning.

eriknelson:
It feels bad to me that the kebab is hidden by default, is that a deliberate design decision? Hiding it doesn't signal to me I can interact with it to get some pretty important features, I would want that kebab displayed by default.

eriknelson (commented):
Thanks again for this Mike I really appreciate it, it feels a lot better. I just want to see how everyone feels about that kebab menu being hidden by default, otherwise it LGTM.

eriknelson (approved):

PR #1015, opened 2020-09-11
Early discovery service cert acceptance

### LINKED ISSUE #1004: Must add discosvc call during app init so user can accept cert
Important: Viewer will error out with network error if you try to view the
tree prior to running through the migplan wizard (which is the way that
you indirectly accept the discovery service network cert). We discussed in
the past that the discovery service cert issue would be solved naturally
due to non-admin flow, but now that non-admin has taken a back seat, we
need to come up with a plan so that people can accept this cert early in
the flow. A dummy request on app startup may be the easiest solution and
best user experience since it will be just another cert acceptance in the
line of them.

eriknelson:
Closes #1004 

It's been agreed that it's a better experience for users to accept the
discovery service cert as part of app initialization with the rest of
the certificates that have to be accepted instead of having it interrupt
the middle of plan creation.

Because the debug tree feature also depends on this cert acceptance,
it's a good time to bump the initial call to app initialization. This PR
introduces the check during app init, and makes sure to wait until the
actual host MigCluster has been loaded, since it's actually possible the
name of that cluster will differ from the default value.

### DIFF
--- a/src/app/AppComponent.tsx
+++ b/src/app/AppComponent.tsx
@@ -46,7 +46,6 @@ interface IProps {
   updateTokens: (updatedTokens) => void;
   clusterList: ICluster[];
   debugTree: IDebugTreeNode;
-  initializeDiscoveryService: () => void;
 }
 
 const AppComponent: React.SFC<IProps> = ({
@@ -70,13 +69,7 @@ const AppComponent: React.SFC<IProps> = ({
   updateTokens,
   clusterList,
   debugTree,
-  initializeDiscoveryService,
 }) => {
-  // Early prompt for discovery service cert acceptace if it hasn't already been accepted
-  useEffect(() => {
-    initializeDiscoveryService();
-  }, []);
-
   const handlePlanPoll = (response) => {
     if (response) {
       updatePlans(response.updatedPlans);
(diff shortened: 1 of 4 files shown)


src/app/AppComponent.tsx
```diff
@@ -69,7 +70,13 @@ const AppComponent: React.SFC<IProps> = ({
   updateTokens,
   clusterList,
   debugTree,
+  initializeDiscoveryService,
 }) => {
+  // Early prompt for discovery service cert acceptace if it hasn't already been accepted
+  useEffect(() => {
+    initializeDiscoveryService();
+  }, []);
+
```
mturley, line 79: @eriknelson, not sure if it matters, but is there a reason you put this here and not in the `appStarted` saga?

mturley: Ah I see, it's probably to make sure resources are loaded first

mturley:
@eriknelson were you going to push one more commit to this or do you want to merge it as-is and follow up?

ibolton336:
Successfully tested on a fresh 1.3 env this morning. LGTM!

src/app/AppComponent.tsx
ibolton336: Played around with moving this call to the appStarted saga & looks like everything is still working as expected.

ibolton336 (approved):
LGTM

PR #1018, opened 2020-09-15
Fix locked in selection bug

### LINKED ISSUE #765: (bug): unable to deselect src/target cluster selection
If a user accidentally selects a cluster within the src/target dropdowns inside the select resources step, the app will not allow a change to be made. 

dropdown_bug

ibolton336:
Closes #765 
image


mturley (approved):
LGTM

PR #1022, opened 2020-09-23
Add build config for logo

ibolton336:
This PR addresses the issue https://bugzilla.redhat.com/show_bug.cgi?id=1881970

Now we can add a .env file with the contents 
```
BRAND_TYPE=RedHat
```
at build time to override the default Konveyor brand setting. This will allow us to display the redhat logo downstream.

Note: As of January 2020, ES2020 supports Promise.allSettled out of the box removing the need for the Q library. This library was causing some dependency clashes with the dotenv webpack plugin, so it is removed as a part of this PR.


config/webpack.dev.js
```diff
@@ -31,10 +31,11 @@ const PORT = process.env.PORT || localConfig.devServerPort;
 const EXPRESS_PORT = process.env.EXPRESS_PORT || 9001;
 
 const plugins = [
-  new webpack.NoEmitOnErrorsPlugin(),
```
mturley, line 34: What's the reason for removing these other plugins?

ibolton336: VScode had a line crossed through them saying they were deprecated. Cant find anything to support that. I will add them back in.

ibolton336: image

mturley: oh interesting ok. 👍

mturley (approved):
LGTM 👍

eriknelson:
Trying to confirm this actually works building an image and deploying, will ack and merge as soon as I can do that

eriknelson (approved):
Confirmed the fix (using the correct env var helps..), LGTM and I've communicated the fix with the downstream guys, thank you!

PR #1024, opened 2020-09-29
Revert to using Q.allSettled again

eriknelson:
Production image uses node 10.x, which does not yet support
Promise.allSettled


mturley (approved):
LGTM

PR #1026, opened 2020-09-30
Update title text to MTC branding

### LINKED ISSUE #1025: Tab is still labeled Cluster Application Migration
image

ibolton336:
Closes #1025 

image

### DIFF
--- a/public/index.ejs
+++ b/public/index.ejs
@@ -1,23 +1,20 @@
 <!DOCTYPE html>
 <html lang="en-US">
+  <head>
+    <meta charset="utf-8" />
+    <title>Migration toolkit for Containers</title>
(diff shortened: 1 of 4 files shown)


public/index.ejs
```diff
@@ -1,23 +1,20 @@
 <!DOCTYPE html>
 <html lang="en-US">
+  <head>
+    <meta charset="utf-8" />
+    <title>Migration toolkit for Containers</title>
```
eriknelson, line 5: Should be "Migration Toolkit" uppercase T

eriknelson (changes requested):

public/index.ejs
ibolton336: updated

eriknelson (approved):

PR #1034, opened 2020-10-23
Skip MigPlan readiness check at PV discovery

djwhatle:
This is a temporary measure that should be reversed after changes
are made to mig-controller so that MigPlan readiness does not
depend on Migration Registry readiness before the registry is needed.


ibolton336 (approved):
LGTM!

PR #1052, opened 2020-10-30
Update react-json-view

eriknelson:


ibolton336 (changes requested):
Should there be a package.json change for this? Only seeing the yarn.lock updated.

ibolton336 (approved):
LGTM- Was confused about how yarn.lock works.

eriknelson:
Closing this one, there's nothing we can do about it since we're on the latest version of our direct dep and the problematic dep is many layers down:    `- Hoisted from "react-json-view#flux#fbjs#isomorphic-fetch#node-fetch"`

PR #1058, opened 2020-11-09
Add support for direct migrations

ibolton336:
-- Adds exposedRegistryURL field in addEditClusterForm
-- Adds a new MigrationOptions step in the wizard which allows Read-Only view of registry paths for selected clusters, check boxes for image and volume migration type selection, and visual indicators for if nthe registry path fields are available on the cluster status. 

**This has not been tested with the controller code.


src/app/cluster/duck/sagas.ts
```diff
@@ -327,7 +328,18 @@ function* updateClusterRequest(action) {
   const currentCABundle = currentCluster.MigCluster.spec.caBundle;
   const caBundleUpdated = clusterValues.caBundle !== currentCABundle;
 
-  if (!urlUpdated && !tokenUpdated && !azureUpdated && !requireSSLUpdated && !caBundleUpdated) {
+  const currentExposedRegistryPath = currentCluster.MigCluster.spec.exposedRegistryPath;
```
pranavgaikwad, line 331: Just curious, what happens when `exposedRegistryPath` is undefined in the cluster spec

pranavgaikwad (commented):
Code looks good. Giving it a try.

src/app/cluster/duck/sagas.ts
sseago: It's a string value, so not defining it should be equivalent to defining it to be an empty string. It is expected to be empty for any cluster where there is no exposed path -- the host cluster in most cases, since for the host cluster we can use the internal path, and remote clusters where the user isn't using direct image migration, since there's no need to expose the registry in that case.

As part of MigPlan src/dest cluster validation, if the plan is using direct image migration and there is no registry path available (internal path for host cluster, exposed registry path otherwise), a critical condition is added to the migplan.

pranavgaikwad (approved):
I gave this a try. Looks good

PR #1060, opened 2020-11-13
Fix direct migration bug

ibolton336:
-- Follow up PR to address a rendering issue on migrations option plan wizard step. 
-- Also removes field validation for the cluster registry path field
-- Ensures changes to plan spec are applied to the patch when updating the plan.


eriknelson (approved):
Sounds like the controller guys are having success with this, we'll just need to add in some kind of a registry route validation. Will file and issue and we'll fix before release.

sseago:
@eriknelson So in terms of registry route validation, instead of validating that it's a URL (since it is not), it's probably better to validate that it's either a valid hostname or a host and port -- i.e. "image-registry-openshift-image-registry.apps.cluster-foo.mycluster.mg.example.com" is valid, as is "image-registry.openshift-image-registry.svc:5000"

PR #1061, opened 2020-11-13
Fix cell widths to prevent flicker when updating

ibolton336:
- Fixes a minor issue with progress reporting where the table is flickering on render due to dynamic sizing of cells.


vconzola:
@ibolton336 Is this for the migration drill down page that Pranav demoed for progress reporting in the sprint 192 review? If so, thank you. That was driving me crazy. :-) If not, we need to fix that one too.

pranavgaikwad (changes requested):
LGTM

pranavgaikwad (approved):

PR #1064, opened 2020-11-17
[WIP] Pipeline updates for migration succeded with warnings

### LINKED ISSUE #1069: [MIG-119] Add first pass migration error / warning condition handling
Today the controller will write N warning conditions to the migmigration in the event of Velero partial failures during specific phases of a migration. We do not have enough visibility into the error back from velero to surface the root cause, but we do have the backup/restore name. Additionally, `CriticalCondition` conditions are written to the migmigration when critical dependencies are not met such as the absence of healthy image registries. This will cause the migmigration to halt until the dependency requirement is satisfied.

**TODO**

* As part of the UI's polling of the migmigration, for each warning message a persistent toast should be shown with the contents of the warning message. A red critical toast should be shown for the CriticalCondition with the message.
* A warning icon and/or a critical icon should show next to the pipeline control with a popover containing the messages as well. This is to still provide a way to surface these warnings even if a user has reloaded the page and toasts are not present.
* On the drilldown screen with the steps, if warnings are present or a critical condition is present, a patternfly alert should be shown above the table but below the header that also contains the messages. The warning alert should be separate from the critical condition, and both should be shown if they are both present.

**Additional Questions**
> What do we show when there are no warning/error conditions?

No icons, no alerts

> Do we want to surface this list on the details page alone or make it available on the top level migration list page also? Maybe the popover from the pipeline summary will be enough here

Popover triggered by clickable icons is enough for now, we'll circle back and can improve after this is implemented if we decide it's not.

> Are the warnings closable? Do we want some sort of show/hide button for the conditions list?

Let's ignore this for now and once it's implemented we can file issue to decide.

pranavgaikwad:
Closes #1069 

This PR 
* Adds a warning icon with popover on the Pipeline summary view to indicate migration succeded with warnings:

Screenshot from 2020-11-19 11-53-28

* Adds Alert on the migration details page to indicate the warnings and prints warnings in the alert

Screenshot from 2020-11-19 11-53-36

Remaining work:
* Show similar Alert for Critical conditions raised during in-progress migration
* Adjust the distance between warning icon and Pipeline summary view as per Mockups

### DIFF
--- a/src/app/home/pages/PlansPage/components/MigrationsTable.tsx
+++ b/src/app/home/pages/PlansPage/components/MigrationsTable.tsx
@@ -6,15 +6,15 @@ import {
   Bullseye,
   EmptyState,
   Title,
-  Progress,
-  ProgressSize,
-  ProgressVariant,
   Spinner,
   Level,
   LevelItem,
   Pagination,
 } from '@patternfly/react-core';
 import spacing from '@patternfly/react-styles/css/utilities/Spacing/spacing';
+import { Popover, PopoverPosition } from '@patternfly/react-core';
+import WarningTriangleIcon from '@patternfly/react-icons/dist/js/icons/warning-triangle-icon';
+
 const styles = require('./MigrationsTable.module');
 
 import { IMigration } from '../../../../plan/duck/types';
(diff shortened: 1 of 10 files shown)


eriknelson:
@pranavgaikwad @ibolton336 let's make sure this PR implements all the TODOs in #1069

PR #1068, opened 2020-11-18
Fixes app not redirecting when token is expired

ibolton336:
The expiry handler is not triggering a redirect when the token is expired resulting in the page hanging in a broken state when a token is invalid or expired. This PR fixes this by forcing a redirect.


ibolton336:
image

ibolton336:
When debugging the node server, it is evident that the current history.push is not reaching the node middleware. Changing this to a hard refresh using the window object allows node to pick it up.

pranavgaikwad (approved):
ACK

PR #1071, opened 2020-11-19
Disable rollback button after completion, show "Rollback Succeeded" text

djwhatle:
image
image

### DIFF
--- a/src/app/home/pages/PlansPage/components/PlanActions.tsx
+++ b/src/app/home/pages/PlansPage/components/PlanActions.tsx
@@ -28,6 +28,7 @@ const PlanActions = ({ plan, history }) => {
     hasErrorCondition,
     hasRunningMigrations,
     hasAttemptedMigration,
+    hasSucceededRollback,
     finalMigrationComplete,
     isPlanLocked,
   } = plan.PlanStatus;
@@ -97,6 +98,7 @@ const PlanActions = ({ plan, history }) => {
         !hasReadyCondition ||
         hasErrorCondition ||
         hasRunningMigrations ||
+        hasSucceededRollback ||
         isPlanLocked
       }
     >
--- a/src/app/home/pages/PlansPage/helpers.ts
+++ b/src/app/home/pages/PlansPage/helpers.ts
@@ -34,9 +34,9 @@ export const getPlanStatusText = (plan: IPlan) => {
   if (hasCancelingCondition) return `Canceling ${latestType}`;
   if (hasRunningMigrations) return `${latestType} Running`;
   if (hasCanceledCondition) return `${latestType} canceled`;
-  if (hasSucceededRollback) return 'Rollback Succeeded';
-  if (hasSucceededMigration) return 'Migration Succeeded';
-  if (hasSucceededStage) return 'Stage Succeeded';
+  if (hasSucceededRollback) return 'Rollback succeeded';
+  if (hasSucceededMigration) return 'Migration succeeded';
+  if (hasSucceededStage) return 'Stage succeeded';
   if (hasReadyCondition) return 'Ready';
   if (isPlanLocked) return 'Pending';
   return 'Waiting for status...';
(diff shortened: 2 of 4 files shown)


vconzola (changes requested):
Tiny nit. "Succeeded" should be "succeeded". We use sentence case everywhere except in the navigation. One other thing...once rollback is complete Stage, Migrate and Edit should become enabled again.

ibolton336:
I wonder what we can show in the pipeline summary view for rollbacks. 
image

vconzola:
> I wonder what we can show in the pipeline summary view for rollbacks.
> image
> 
> Maybe a spinner that uses the isRollbackPolling boolean as a source?

Does rollback have steps and phases like other migrations? If so, we could use the same pipeline view.

ibolton336:
@vconzola I do not think that is implmented yet. The next priority is getting pipeline steps for Direct Migrations. Not sure where that would fall on the roadmap.

vconzola:
@ibolton336 @djwhatle If we don't have any pipeline information (step/phase) during rollback, then I agree that the best thing to do is to show the spinner icon with the the text "In progress". Unless, of course, we get progress status in which case we should show a progress indicator. But I'm guessing we don't get that either?

djwhatle:
> Tiny nit. "Succeeded" should be "succeeded". We use sentence case everywhere except in the navigation. One other thing...
Let me see what I can do. I have carried over the existing convention that is implemented in the code. What you're saying is that we should change all occurrences of this in the UI.

> once rollback is complete Stage, Migrate and Edit should become enabled again.
This is not feasible for 1.4.0 because it would require significant controller + UI changes and we are at code freeze.

djwhatle:
@ibolton336 @vconzola I agree with you that we should fix the pipeline view for rollbacks (tracking in [812](https://github.com/konveyor/mig-controller/issues/812) but that is a separate concern from this PR in my opinion, so I will fix the capitalization error that Vince mentioned here but leave the other problems to be solved in a different PR.

djwhatle:
@ibolton336 @vconzola changed the case as suggested.

ibolton336:
Looks great - thanks @djwhatle

ibolton336 (approved):
LGTM 💯

vconzola:
@djwhatle If we can't re-enable the Stage, Migrate, and Edit buttons for a plan that's been rolled back what's the status of that plan? Is it just orphaned? I'm not sure why someone would rollback a plan, but can assume it's because something was wrong in the plan configuration. Editing and retrying the plan would seem like obvious next steps. What am I missing in my logic?

djwhatle:
> @djwhatle If we can't re-enable the Stage, Migrate, and Edit buttons for a plan that's been rolled back what's the status of that plan? Is it just orphaned? I'm not sure why someone would rollback a plan, but can assume it's because something was wrong in the plan configuration. Editing and retrying the plan would seem like obvious next steps. What am I missing in my logic?

You could rollback because something was wrong in plan config, or because the migration was unsuccessful and changes need to be reverted. Right now, if you complete a rollback the plan will be in a terminal state, meaning that you should delete the plan and start over.

vconzola:
@djwhatle Thanks for the detailed explanation. I'll talk to @eriknelson about getting these Retry changes into a post 1.4 release because having to recreate a plan every time is kinda ug-ly from a UX standpoint.

apinnick:
> Tiny nit. "Succeeded" should be "succeeded". We use sentence case everywhere except in the navigation. One other 

@vconzola Patternfly4 updated their guidelines. Now they use sentence case in navigation. https://www.patternfly.org/v4/ux-writing/capitalization

vconzola:
@apinnick General PF guidance is sentence case in navigation, however, cloud.redhat.com is an exception in that they're using title case for nav. And since all of our migration apps will eventually live on c.rh.c, I want to follow the same pattern.

apinnick:
> @apinnick General PF guidance is sentence case in navigation, however, cloud.redhat.com is an exception in that they're using title case for nav. And since all of our migration apps will eventually live on c.rh.c, I want to follow the same pattern.

Life is never simple. :-) 
Good to know that cloud portal is an exception.

PR #1073, opened 2020-11-23
DVM Progress

### LINKED ISSUE #1070: [MIG-119] Implement first pass DirectVolumeMigration (DVM) Progress
Direct Image Migration progress will not be available to us as part of 1.4.0. We are not expecting to display anything in the DIM row. For DVM, there is some amount of time between when the step has started and when the controller has started to write volume progress to the CR so that we can render the volume progress. During this time, instead of displaying the "View Details" link, we should replace with the DVM's "phaseDescription", a human readable string that describes what the DVM controller is actually working on at that moment. Once the volume progress details are available, we should display the "View Details" link so that the user can view individual volume progress on the 2nd drilldown level (just like a restic copy).

pranavgaikwad:
Closes #1070 

This PR : 
* Adds logic to parse DVM progress messages
Screenshot from 2020-11-24 12-39-07

* Adds support for marking Migration steps 'skipped'
Screenshot from 2020-11-23 16-26-31

Screenshot from 2020-11-24 12-46-28

* Fixes helper logic to discard un-parseable progress messages for sanity
* De-duplicates logic for progress message parsing
* Fixes breadcrumbs not to reload the app

### DIFF
--- a/src/app/home/pages/PlansPage/components/MigrationsTable.tsx
+++ b/src/app/home/pages/PlansPage/components/MigrationsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useEffect } from 'react';
 import { Table, TableBody, TableHeader, cellWidth, sortable, IRow } from '@patternfly/react-table';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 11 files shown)


ibolton336 (approved):
LGTM!

PR #1074, opened 2020-11-24
Initial GH actions workflow for ci

ibolton336:


eriknelson:
@ibolton336 what's the background on this, are we trying to move away from Travis? (+1 if that's the case, it has not been very responsive lately).

ibolton336:
@eriknelson Yeah we were having to wait hours for travis builds so swapped over to github actions. I think it offers the same functionality & it is free for public repos. We may need to double back over these changes just to confirm everything is working the way we want.

PR #1080, opened 2020-12-03
Fix an issue with direct migration being set to enabled when not available 

### LINKED ISSUE #1079: Bug 1904165: PV discovery screen raises error in mig-ui with new direct migration setup
**Describe the bug**
https://bugzilla.redhat.com/show_bug.cgi?id=1904165
![image](https://user-images.githubusercontent.com/7576968/101070278-209e1000-3569-11eb-98af-8dd9a7525aad.png)


**To Reproduce**
1. Select namespaces to migrate in mig-ui
2. Proceed to PV discovery step
3. UI will raise error condition warning "Direct image migration is selected and the source cluster openshift-migration/cluster-3 is missing a configured Registry Path"
4. Strangely, the issue seems to be resolved eventually. I checked the spec of the migplan and saw this:
```
  spec:                                                                                                                                                                                                   
    destMigClusterRef:                                                                                                                                                                                    
      name: host                                                                                                                                                                                          
      namespace: openshift-migration                                                                                                                                                                      
    indirectImageMigration: true                                                                                                                                                                          
    indirectVolumeMigration: true
```
My thought is that the initial migplan created by mig-ui is missing the `indirectImageMigration` and `indirectVolumeMigration` settings so the migplan is defaulting to use direct migration.

**Expected behavior**
The migration UI should not throw an error during PV discovery when the user hasn't even had a chance to select indirect/direct migration yet.

This could be solved on controller side by flipping the default and renaming the `directImageMigration` vars, or it could be fixed in mig-ui by making sure the first version of a migplan that is created has `indirectImageMigration` and `indirectVolumeMigration` both set to true.

**Additional context**
I'll file a BZ for this as well.

ibolton336:
Closes #1079

### DIFF
--- a/src/client/resources/conversions.ts
+++ b/src/client/resources/conversions.ts
@@ -521,6 +521,11 @@ export function createInitialMigPlan(
         namespace,
       },
       namespaces,
+      /** Set the initial value for indirect migration to true until the user navigates to the selection screen 
+        where the spec fields will be updated to represent availablility/user selection.
+      **/
+      indirectVolumeMigration: true,
+      indirectImageMigration: true,
     },
   };
 }


djwhatle:
I still see the issue 
image

djwhatle (approved):
This worked well for me and resolved the error I posted a screenshot of. 

It's worth noting that the way this solves the problem is by initially setting the migration plan up with indirectImageMigration: true and indirectPVMigration: true. Once you get to the step where the user has the option to choose, these initial settings may be overridden. I think this is a suitable solution.

PR #1081, opened 2020-12-07
Fix for empty ns bug

ibolton336:
Adds a fix for BZ [1904414](https://bugzilla.redhat.com/show_bug.cgi?id=1904414 )
- This bug occurs when the pv list is empty. This is fixed by checking for persistentVolumes on the spec on the options form page.


pranavgaikwad (approved):
ACK

djwhatle (approved):
VISACK

PR #1083, opened 2020-12-08
Update storage form to show proper type when editing for azure/gcp

ibolton336:
Fixes [BZ 1882061](https://bugzilla.redhat.com/show_bug.cgi?id=1882061)

- When editing, we werent setting the provider type for the storage before rendering the storage form. This PR fixes this for GCP/Azure provider types.


pranavgaikwad (approved):
ACK

PR #1086, opened 2020-12-10
add check for migration.status.pipeline before setting in selector

### LINKED ISSUE #1085: When a migration has no status information, pipeline set fails
This occurs when the controller is in a broken state. The selector is attempting to set some derived state on the UI copy of the migration, but fails because there is no status present.

ibolton336:
Closes #1085


pranavgaikwad (approved):
ACK

eriknelson:
Any BZ associated with this one?

ibolton336:
Nope this was one Scott brought up in the channel yesterday.

PR #1087, opened 2020-12-10
Remove local k8s client code in favor of shared lib-ui 

ibolton336:
- Removes local k8s client code & favors imported shared code from the lib-ui repo.
- Stub out reverse proxy work to solve cors issues


mturley:
Nice! At some point soon I'd love to move some of the TS types from virt-ui into the client itself to reuse here

PR #1091, opened 2020-12-15
[BZ 1877439] - Check location for existing secret instead of assuming openshift-config

ibolton336:
Closes [bz 1877439](https://bugzilla.redhat.com/show_bug.cgi?id=1877439)


djwhatle (approved):
VISACK

Issue #920, opened 2020-06-17
When creating a new hook PostRestore needs to be selected twice in order to proceed


jmontleon:
$subject

PR #1099, opened 2021-01-08
Add global hooks

ibolton336:
-  Make hooks 1st class objects with CRUD operations from main nav
-  Allow global hooks to be associated with a plan at plan creation time & during edit


src/app/home/pages/HooksPage/components/AddEditClusterModal/AddEditClusterForm.tsx
```diff
@@ -0,0 +1,387 @@
+import React, { useState } from 'react';
```
pranavgaikwad, line 1: I couldn't find any cluster actions on Hooks page. I didnt understand the usage of the Cluster Edit modals and forms in the context of hooks.

pranavgaikwad (dismissed):
Overall, the code looks good to me. Works as expected. One small question about usage of cluster edit forms in the context of Hooks.

src/app/home/pages/HooksPage/components/AddEditClusterModal/AddEditClusterForm.tsx
ibolton336: Good catch @pranavgaikwad - removed these files.

pranavgaikwad: Ahh I see. They were pasted from the directory.

ibolton336: @pranavgaikwad everything seems to be working after removing those files. Yeah i forgot to remove those after planning to use them as templates.

pranavgaikwad (approved):
ACK

ibolton336:
Closes #920

PR #1101, opened 2021-01-12
Bug 1915702: Add check for empty step name when cancelling rollback

ibolton336:
This fixed a bug causing a UI crash when a rollback was cancelled. There is a temporary controller state causing the reason field to be blank in status.


ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=1915702

eriknelson (approved):

PR #1102, opened 2021-01-21
Remove stage/migration/rollback polling

### LINKED ISSUE #1100: change the warning condition message on migmigration
For 1.4.0 the UI has a feature that displays the warning condition from the migmigration CR in to the UI. 

The warning however says that Stage operation is complete even when the operation is running as seen in the following screenshot. @pranavgaikwad confirmed this is indeed a bug

image

ibolton336:
Closes #1100 

- This PR removes redundant polling for migmigration objects. This is old code that was made obsolete by the recent improvements in migration details status/progress reporting.


pranavgaikwad (approved):
ACK

PR #1103, opened 2021-01-22
Check for null conditions array when looking for latest rollback

ibolton336:
- Fix the UI blowing up when conditions array is temporarily null


pranavgaikwad (approved):
ACK

PR #1110, opened 2021-01-28
Bug fix - Fix small regression with warning condition being hidden on migmigration

ibolton336:
This PR moves the check for warnings above the check for succeeded migrations to avoid swallowing the check for warnings on a migmigration. Currently if a migmigration has a warning, it is not displayed in the UI if there is a succeeded condition present. 

Fixes regression introduced with https://github.com/konveyor/mig-ui/commit/1ca0bd30e602ad3147fddb63901866c2348e24fa

### DIFF
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -395,7 +395,6 @@ const getPlansWithStatus = createSelector([getPlansWithPlanStatus], (plans) => {
       return status;
     }
 
-
     return status;
   };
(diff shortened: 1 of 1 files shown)


pranavgaikwad (approved):
ACK

eriknelson:
I think we want this in 1.4.0 as John pointed out, this could be a problem
for our 1.4.0 labs in Feb? I don't have a BZ for this yet but it might
still be good for me to write one up so it gets verified. I'll do that this
morning

PR #1114, opened 2021-01-29
Bug 1920664: Add plan level warning conditions

ibolton336:
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1920664 
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1921898
image

image

image

### DIFF
--- a/src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
+++ b/src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
@@ -23,6 +23,8 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
     hasNotReadyCondition,
     hasSucceededStage,
     hasSucceededMigration,
+    hasSucceededMigrationWithWarnings,
+    hasSucceededStageWithWarnings,
     isPlanLocked,
     hasConflictCondition,
     latestIsFailed,
@@ -60,7 +62,9 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
     return <Spinner size="md" />;
   } else if (
     (hasSucceededMigration && hasWarnCondition) ||
-    (hasSucceededStage && hasWarnCondition)
+    (hasSucceededStage && hasWarnCondition) ||
+    hasSucceededMigrationWithWarnings ||
+    hasSucceededStageWithWarnings
   ) {
     return (
       <span className={`pf-c-icon pf-m-warning`}>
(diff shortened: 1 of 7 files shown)


eriknelson:
This is exposing a gap in our BZ automation -- a fix that applies to more
than one BZ... Thinking maybe just choose one to preserve the "Bug XXX:"
syntax so the bot picks it up, and manually shepherd the other one through
POST and MODIFIED?

src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
```diff
@@ -50,6 +53,20 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
         </span>
       </Popover>
     );
+  } else if (hasDVMBlockedCondition) {
+    return (
+      <Popover
+        position={PopoverPosition.top}
+        bodyContent="Some or all rsync transfer routes have failed to be admitted within 3 mins on destination cluster. Errors: [no status condition available for the route]"
```
pranavgaikwad, line 60: Can we take the body content from the condition message instead?

pranavgaikwad (changes requested):

vconzola:
A detailed warning message like this shouldn't be displayed at the plan level. It should be at the migration or step/phase level. The plan should show a finite (and small) number of states and guide the user to drill down for further details. I suggest for the plan level we leave the Last state as “Running” with a warning icon. Then for the migration status say something like “Paused - waiting for route to be admitted” above the pipeline. The full verbose message above should be shown as phase feedback.

src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
```diff
@@ -50,6 +53,25 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
         </span>
       </Popover>
     );
+  } else if (hasDVMBlockedCondition) {
+    const DVMBCondition: ICondition = plan.Migrations.reduce(
```
ibolton336, line 57: @pranavgaikwad Ideally this logic should exist in a selector but felt beyond the scope of this PR to try to refactor. I would like to build a more refined condition sorting mechanism on the UI side that can put conditions into info/warning/error buckets with prioritization. This will fall into the bucket of work for bringing in a multi-condition view component like we have on the plan validation screen. Similar to what has been done on mtv : [](https://github.com/konveyor/forklift-ui/blob/main/src/app/common/components/StatusCondition.tsx)

eriknelson (changes requested):
See @vconzola feedback

src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
ibolton336: removed the above logic in favor of encouraging the user to drill down to migration details view.

ibolton336:
image

image

image


Updated

pranavgaikwad (approved):
ACK

PR #1119, opened 2021-02-10
[Bug fix] - Resume polling bug & cert error flow enhancements

### LINKED ISSUE #1118: Error modal not starting polling when "Close [x] icon" is clicked
@rayfordj Raised an issue around behavior inconsistency in the Error modal. We should resume polling when the [X] button is pressed.

### LINKED ISSUE #1135: UX improvements for cert error condition
After a user installs MTC, they are presented with this screen when navigating to the UI route:

image

If a user clicks [Refresh Certificate], they are redirected to a page where presented with the URL to click the .well-known/oauth- link and accept cert.

Once they have accepted the certs, going back & refreshing (reloading the ui) will no longer present the modal with error.

This could be improved by adding an empty state or some better UX. 

There is one caveat with the error we have available to us to decide if there is a certificate error present. The error is the same for any "Network Error".

ibolton336:
Close #1118 
Close #1135 
cert-error


mturley:
@ibolton336 it might be a neat idea to introduce the `usePausedPollingEffect` hook from MTV into MTC. We could do it without changing much of the actual polling implementation, we would just have a `useEffect` in the hook calling the `stopAllDefaultPolling` and `startAllDefaultPolling` actions on mount/unmount. That way you don't have to remember where to call them imperatively, just call the hook in any component that needs to have polling off while it's mounted.

ibolton336:
@mturley For sure that is a great call. I will open an issue to capture this enhancement. I am a big fan of how clean that implementation is.

ibolton336:
image

PR #1123, opened 2021-02-17
Bug 1926110: proxy fix

ibolton336:
Closes https://bugzilla.redhat.com/show_bug.cgi?id=1926110 

Need to work with @jmontleon to integrate with operator env variable for proxy URL

### DIFF
--- a/deploy/main.js
+++ b/deploy/main.js
@@ -54,8 +54,9 @@ app.get('/login/callback', async (req, res, next) => {
     redirect_uri: migMeta.oauth.redirectUrl,
   };
   try {
+    const proxyString = process.env['HTTP_PROXY'] || '';
     const httpOptions = {
-      agent: new HttpsProxyAgent(migMeta.proxyString),
+      agent: new HttpsProxyAgent(proxyString),
     };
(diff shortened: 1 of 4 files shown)


deploy/main.js
```diff
@@ -49,11 +51,15 @@ app.get('/login/callback', async (req, res, next) => {
   const { code } = req.query;
   const options = {
     code,
-    redirect_uri: migMeta.oauth.redirectUri,
+    redirect_uri: migMeta.oauth.redirectUrl,
   };
   try {
+    const proxyString = process.env['HTTPS_PROXY'] || process.env['HTTP_PROXY'] || '';
+    const httpOptions = {
+      agent: new HttpsProxyAgent(proxyString),
```
pranavgaikwad, line 59: When proxy string is not set or set to "", is the Proxy ignored? I assume when the user doesn't set it, operator always sets it as empty string.

pranavgaikwad (approved):
VISIACK. added one question for clarification.

deploy/main.js
ibolton336: @pranavgaikwad Updated to bypass this setting if the proxy is not set in the env vars.

PR #1124, opened 2021-02-22
Bug 1920452: Do not show loading screen when pipeline is not available

pranavgaikwad:
This PR updates migration details page to check whether pipeline field is available on a MigMigration. If it's not found but MigMigration is complete, it is likely that the MigMigration was created before MTC 1.4.0, when pipeline was not available on MigMigration status.


ibolton336:
Tested & LGTM! @pranavgaikwad Thank you & nice work.

ibolton336 (approved):

PR #1125, opened 2021-02-24
Bug 1926774: filter out move & snapshot pv selections

ibolton336:
closes https://bugzilla.redhat.com/show_bug.cgi?id=1926774


pranavgaikwad (approved):
VISIACK

PR #1128, opened 2021-03-02
Global hooks bug fix - Exposes src/target selection in add/edit global hook form 

ibolton336:
Closes https://github.com/konveyor/mig-ui/issues/1121 

image

Disabled src/target selection for adding an existing hook to a plan or editing an existing hook inside a plan. 

image

SA name & namespace fields conditionally render based on src/destination selection. 
image

### DIFF
--- a/src/app/common/duck/utils.ts
+++ b/src/app/common/duck/utils.ts
@@ -30,6 +30,7 @@ export const isSelfSignedCertError = (err) => {
 };
 
 export const isTimeoutError = (err) => {
+  //TODO: We should do some type checking here. Have seen the toJSON() conversion fail. Maybe use stringify.
   const e = err.toJSON();
   return e.code && e.code === 206;
 };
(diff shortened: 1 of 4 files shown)


vconzola (commented):
@ibolton336 Why are we adding the source/target selection to global hooks? That should be part of the plan CR, not the hook CR. I think the user experience should be driving the backend design here, not vice versa. @eriknelson interested in your thoughts.

eriknelson:
1.4.2 ships in 8 days; we need a solution for that timeframe so hooks are functional.

Post-1.4.2 we should lead with the desired UX and prioritize it within the broader context of our roadmap's goals.

pranavgaikwad (approved):
ACK, 
I tested these scenarios and worked as expected: 
1. Adding Hook from Hook page, selecting to run it on Target / Source cluster
2. Selecting an existing Hook in MigPlan wizard, make sure the checkboxes are disabled
3. Create a new Hook from MigPlan wizard

ibolton336:
@pranavgaikwad Thanks for all the reviews :+1:

PR #1129, opened 2021-03-04
Bug 1913075: Strip http(s) from registry path & update validations/copy

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=1913075


pranavgaikwad (approved):
LGTM

PR #1132, opened 2021-03-10
Add case for squid proxy failure & display relevant error page

ibolton336:
This PR handles a case where a squid proxy server has hit a TLS/SSL error. By catching this and surfacing to the user, this will show the user which SSL cert in the chain is causing the request to break down. 

For now the solution is to add this line to the squid conf which turns off SSL:

``` sslproxy_cert_error allow all ``` 

Going forward, we could potentially look into configuring our UI to handle SSL which would eliminate CORS errors & manual cert acceptance.
This will require a few changes:
- on the operator side, will need to configure local reverse proxy urls. Will also need to provide Node with the CA cert for the cluster API.
- on UI side configure the reverse proxy: https://github.com/konveyor/mig-ui/pull/1089


pranavgaikwad (commented):
Just wanted to make sure that 503 doesn't appear for any other errors. Even if it does, it'd be OK to misrepresent it as proxy error if its an extreme edge case.

ibolton336:
@pranavgaikwad Good call - I think it will be ok to merge since it will just push the error data into a rendered html page no matter what the context of the error is.

PR #1133, opened 2021-03-10
Add polling context and paused effect

ibolton336:
Closes https://github.com/konveyor/mig-ui/issues/1120 

This PR upgrades redux to the latest version. This is a non-breaking change. Version 7.2 introduces the  [redux hooks api](https://react-redux.js.org/api/hooks 
) which unlocks a variety of standard practices that have emerged in the react space. Namely using reusable hooks to handle side effects of an application. 

- Moves all start/stop polling logic out of the AppComponent.tsx file and into the PollingContext.tsx hook file. This new hook file exposes the PollingContextProvider wrapper around our app & the usePausedPollingEffect function. This function automatically handles starting and stopping polling across the app. To use it, simply add usePausedPollingEffect() at the top of a function component that needs to pause polling while it is mounted. 

This is made possible by redux 7.2 hooks api. This now allows us to directly fire redux actions from within a hook.

### DIFF
--- a/deploy/main.js
+++ b/deploy/main.js
@@ -33,18 +33,22 @@ app.get('/login', async (req, res, next) => {
   try {
     const clusterAuth = await getClusterAuth(migMeta);
     const authorizationUri = clusterAuth.authorizeURL({
-      redirect_uri: migMeta.oauth.redirectUrl,
-      // redirect_uri: migMeta.oauth.redirectUri,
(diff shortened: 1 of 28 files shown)


djwhatle (approved):
VISACK, @ibolton336 does this affect all polling in UI (like on the MigPlan wizard?)

PR #1138, opened 2021-03-15
Create debug slice

ibolton336:
Using the reduxjs/toolkit allows us to more easily type our reducers and actions & take advantage of the auto generation of actions to reduce boilerplate. 

https://redux-toolkit.js.org/api/createAction
https://redux-toolkit.js.org/api/createSlice


djwhatle (approved):
VISACK. Will there be any code removal due to this change coming in a later PR? I wasn't sure if the aliases you mentioned were necessary anymore.

ibolton336:
> VISACK. Will there be any code removal due to this change coming in a later PR? I wasn't sure if the aliases you mentioned were necessary anymore.

The PR removes the reducers.ts and actions.ts files as well as the connect functions with the aliases at the bottom of the presentation layer components. Any later changes would be to convert the other sub-reducers to redux "slices".

PR #1143, opened 2021-03-18
Update pf and webpack

ibolton336:
This PR updates PF, webpack & webpack loaders. 

The reason behind this is to pull in the expandable card component from pf & other PF enhancements. When updating PF i became aware of some hard dependencies on the updated version of webpack. The migration to webpack 5 was fairly painless & should give us some additional features around performance & optimization.


djwhatle (approved):
LGTM!

PR #1158, opened 2021-04-05
Add scroll to card effect for view json functionality

ibolton336:
Fixes a poor user experience bug where the json view card would be rendered outside of the viewport. This adds a scroll animation to the viewjson div when the card is initially redered.


ibolton336:
autoscroll

PR #1163, opened 2021-04-09
Enhanced debug fixes

ibolton336:


src/app/debug/duck/selectors.ts
```diff
@@ -146,15 +146,21 @@ const getResourceStatus = (debugRef: IDebugRefRes): IDerivedDebugStatusObject =>
...
       const hasWarning = conditions?.some((c) =>
         checkListContainsString(c.category, warningConditionTypes)
       );
       const hasFailure = conditions?.some((c) =>
         checkListContainsString(c.type, ['InvalidPod', 'InvalidPodRef'])
       );
-      const hasCompleted = false;
-      const hasRunning = false;
+      const hasCompleted = phase === 'Succeeded';
+      const hasRunning = totalProgressPercentage !== '100%';
```
djwhatle, line 163: @ibolton336 it's been my experience that there are cases where totalProgressPercentage may not reach 100% if we don't poll the Rsync Pod logs one last time before it shuts down with Phase=Succeeded. @pranavgaikwad does this sound accurate to you?

I just want to make sure we show "hasCompleted" with greater precedence than "hasRunning".

djwhatle:
@ibolton336 I saw this crash (UI page went blank white) when trying to view the tree for a freshly started stage migration.

image

djwhatle:
@ibolton336 I think it would be pretty cool if the DVMP status showed the progress percent as its status text when it's between 0 and 100 (e.g. when its still running). Do you think that would be hard to add this PR? Otherwise will do it as a follow-on assuming you're ok with the idea.

ibolton336:
I like that idea @djwhatle should be easy enough to add to this PR.

ibolton336:
image

djwhatle (dismissed):
LGTM. I did see the progress jump from 0 to 100%, although I believe that may be a problem on the back-end. I asked QE to verify if this is happening in 1.4.3 images. 

Edit: Found a few more issues, posted below.

djwhatle:
@ibolton336 I think we still need to resolve 

- the issue of dealing with resources are removed from the cluster naturally as part of migration causing 404 in tree view
- DVMP does not show correctly as "Failed" when DVMP.phase=failed
image

ibolton336:
I think I hit everything but the 404 issue. I think I am going to try to address that in another PR @djwhatle

djwhatle (approved):
LGTM! Working out the  404 on deletion of resources in a future PR sounds good. Merge away when you're ready.

PR #1173, opened 2021-04-20
Implement Tree table for debug view

ibolton336:
image

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -69,7 +69,16 @@ const webpackConfig = {
   optimization: {
     removeAvailableModules: false,
     removeEmptyChunks: false,
-    splitChunks: false,
+    splitChunks: {
+      cacheGroups: {
+        default: false,
+        commons: {
+          name: 'commons',
+          chunks: 'all',
+          minChunks: 3,
+        },
+      },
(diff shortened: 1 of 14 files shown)


djwhatle (changes requested):
A few things that jump out to me:

 - Cluster (source, target, host) should be included as a column, this is badly needed for users to be able to know where they can find the resources if they're using the tree view as a jumping off point for more debugging
 
 - If possible to make the tree table more compact, we should. For migrations that aren't just moving one sample app, the tree can expand to include 100s of items so it's useful to have them shown compactly IMO

 - `Kind` is not an accurate reflection of the content of that column currently since `Kind` has meaning at the API level and the stuff we have in that column is the "UI-friendly" versions of kinds instead of the real things. E.g. `MigPlan` is shown as `Plan` in the UI but if you try to run `oc get plan` you're not going to get anything. This could be fixed with a different column name (not sure what name would be best), or a backend change to use real API names in tree view.


_Sample of tree view with a bunch of stuff in it (copying 10 PVs with Direct Volume Migration)_

image

ibolton336:
Closing for now - Unable to get sufficient performance out of the pf table tree variant.

PR #1183, opened 2021-04-23
[MIG-620] Remove progress percentage value from progress message as it is already shown on the progress bar

pranavgaikwad:
Removes redundant progress percentage value from <message> part of a progress string.
Screenshot from 2021-04-23 08-51-55


ibolton336 (approved):
LGTM!

PR #1190, opened 2021-04-29
Use stage, migration, rollback in migmigration names

djwhatle:
This changes the migmigration names generated by the UI to be meaningful with a shorter 5 char UUID for easier readability and human identification.

The chance of collision with a 5 char UUID is 36^5 (1 / 60466176). I think we can live with that for the easier readability benefits it provides.  

image


ibolton336 (approved):
LGTM!

vconzola:
Love this!

PR #1192, opened 2021-04-29
Tree performance enhancements

ibolton336:
Fixes laggy experiecne when the tree view is loaded up with resources.
-Memoize the tree component
- Stop all polling other than plan polling when debug tree view is open


djwhatle (approved):
LGTM

PR #1193, opened 2021-04-30
Style pending debug resources in yellow highlighting

djwhatle:
I realized it's really important to show the user which resources are pending, because pending things, by definition, hold up the migration. 

Right now the blue highlighting kind of overshadows the (unhighlighted) pending resources where the action is happening. I mocked out what it would look like to highlight pending in yellow. 

Please ignore the "active" DVMPs, those need their status fixed, they aren't actually tracking an active transfer yet.

image

### DIFF
--- a/src/app/debug/duck/utils.module.scss
+++ b/src/app/debug/duck/utils.module.scss
@@ -17,9 +17,19 @@
   text-overflow: ellipsis;
 }
 
-.activeHighlight {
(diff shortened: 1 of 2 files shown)


djwhatle:
For comparison's sake, here's the same thing without active highlights

image

djwhatle:
Closing in favor of #1196 to avoid the christmas tree effect 😄

PR #1194, opened 2021-04-30
Change misleading message about route being cause of delay

djwhatle:
The message given when Rsync doesn't start after 10 minutes misleads the user. There are multiple reasons why Rsync could be stuck, and blanketly stating that the route is at fault is not accurate.

**Previous**
```
Paused - waiting for route to be admitted
```

**New**
```
Migration may be stuck.
```

image

In the above case, a bad PV with missing backing store was the cause of failure, completely unrelated to routes.

The next step will be for me to go into the backend and
1. Present this message before 10 minutes of nothing happening.
2. Give the user a call-to-action


djwhatle:
Here's what the current messaging looks like
image

ibolton336:
Looking good!

ibolton336 (approved):
LGTM

PR #1196, opened 2021-05-05
Highlight pending and terminating resources in blue as well

djwhatle:
https://user-images.githubusercontent.com/7576968/117170814-548b9480-ad98-11eb-9df4-38822d55589e.mov


ibolton336:
LGTM @vconzola what do you think?

djwhatle:
My thought here is that "pending" and "terminating" also fall in the bucket of "active" resources. All of these are likely to be interesting to the user.

vconzola:
it is a lot of highlighting, but if it's useful then i'm ok with it. a lighter shade of blue (--pf-global--palette--blue-50) might not be so harsh. it looks like you're currently using (--pf-global--palette--blue-100).

djwhatle:
I'm a fan of the lighter blue! A lot less harsh when there's a bunch of stuff highlighted.

https://user-images.githubusercontent.com/7576968/117880883-9aec6200-b276-11eb-8e89-1e0f5af93582.mov

djwhatle:
@ibolton336 @vconzola updated.

ibolton336:
Tested and LGTM!

ibolton336 (approved):

PR #1205, opened 2021-05-11
Target namespace mapping

ibolton336:
https://user-images.githubusercontent.com/11218376/117891016-2fa98c80-b284-11eb-8609-bb82ff848924.mov


vconzola (changes requested):
Please add the popover on the Target name column header. Also, I can't tell for sure what color it is, but the check icon should be blue. It looks green to me, but I have trouble distinguishing blue and green.

ibolton336:
@vconzola updated

ibolton336:
Screen Shot 2021-05-12 at 11 14 56 AM

vconzola (commented):
LGTM. Thanks Ian.

ibolton336:
Waiting for https://github.com/patternfly/patternfly-react/pull/5770 before merging.

apinnick:
Typo in the first sentence: "names" > "name"

The tooltip belongs to "Target name". Suggest changing to "Target namespace".

Suggested text: "By default, the target namespace has the same name as the source namespace. Click the Edit icon to update the name of the target namespace."

Alternatively, you could also remove the first sentence completely. If the target namespace column is populated, the correspondence between the source and target namespace names should be obvious to the user.

ibolton336:
Screen Shot 2021-05-13 at 9 10 25 AM

@vconzola @apinnick how does this look?

apinnick:
LGTM!

vconzola:
Me too. Thanks @apinnick .

ibolton336:
image

@vconzola After discussing with @sseago , We decided to add validation for the mapped target namespace name input. The validation checks for any duplicate target namespaces that may be mapped & the standard namespace dns1123 validation https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ .Let me know if you think this wording is acceptable.

eriknelson:
How do we know the target namespaces? Is that coming from the discovery
service?

ibolton336:
@eriknelson The target ns is the same as the source by default. We are giving an option to change the target ns which will result in a namespace mapping value of srcNs:targetNs in the yaml for the namespace.

vconzola:
@ibolton336 The validation text LGTM.

eriknelson:
@ibolton336 I'm asking: how are we able to warn the user that there already exists a namespace on the target cluster that conflicts with their entered value? That requires us to have the entire list of namespaces on the target cluster and to verify there is no conflict; the UI doesn't have access to this data as far as I'm aware, unless it's been exposed somehow by something like the discovery service?

ibolton336:
@eriknelson This is to prevent creating a duplicate namespace that can occur when using the text input. They are being warned that they have mapped a namespace to an existing namespace in the src cluster list. When just using the checkbox, they were not able to make this mistake. Maybe there are edge cases where the target namespace has an existing namespace already on the cluster but this doesn't check for that.

eriknelson:
@ibolton336 *OH* I thought this was actually listing all the namespaces that already existed in the target cluster and comparing the new namespaces to that set to see if there is any overlap. We probably want to validate that as well, I'm unsure how we might be able to do that apart from a controller validation on the Plan? Have you discussed that scenario with scott at all?

sseago:
@eriknelson "I'm asking: how are we able to warn the user that there already exists a namespace on the target cluster that conflicts with their entered value? That requires us to have the entire list of namespaces on the target cluster and to verify there is no conflict; the UI doesn't have access to this data as far as I'm aware, unless it's been exposed somehow by something like the discovery service?"

We don't currently warn if the namespace alreay exists. If we decided to do that, this would be independent of the namespace mapping functionality, since even *without* mapping to a new namespace, we don't warn if the namespace already exists.

sseago:
The only namespace duplication that we validate for is if *within a plan* a target namespace is listed twice (two different source namespaces getting mapped to the same target namespace). At the controller/API level, we also validate against source namespace duplication. The user will see a similar validation error if a source namespace is duplicated as will be seen if a destination namespace is duplicated. However, on the UI side, the source namespace duplication can't really happen since the user is just checking boxes from a list of already-existing namespaces on the source cluster.

PR #1213, opened 2021-05-20
Add pending state to PVR/PVB/Backup/Restore

djwhatle:
When these Velero Resources haven't started, show them as "Pending" in the debug view instead of just a blank status.


ibolton336 (approved):
Nice work!

PR #1216, opened 2021-05-25
Discovery service events API integration

ibolton336:
Blocked by: https://github.com/konveyor/mig-controller/pull/1084 
- Makes use of api changes allowing events to be queried by UID
- Uses events to determine if a resource is in a problem state & alerts the user if one or more warning events exist for a resource in the debug tree.


ibolton336:
image

PR #1227, opened 2021-06-09
Move PR-BZ branch_to_release mapping to Org secret

rayfordj:
Revisiting https://github.com/konveyor/bz-github-action/issues/2

@djwhatle , Given the number of incoming PRs expected for 1.5.0, I'd like to enable this as a trial on mig-ui to exercise the change before rolling it out across the remaining repos.  The intention is to reduce the amount of "lift" required to manage the mapping across all repos every time we need to update it.

cc: @eriknelson


djwhatle (approved):
LGTM

PR #1251, opened 2021-06-23
Bug 1975370: Add validation for target ns length

ibolton336:
(Pending BZ) 
Adds length validation for target namespace name when mapping a namespace.

### DIFF
--- a/src/app/common/duck/utils.ts
+++ b/src/app/common/duck/utils.ts
@@ -49,7 +49,7 @@ export const handleSelfSignedCertError = (failedUrl: string, dispatch: any) => {
 
 const testTargetNSName = (value: string) => {
   if (value?.length < 3 || value?.length > 63) {
-    return 'The namespace name can be between 3 and 63 characters long.';
+    return 'The namespace name should be between 3 and 63 characters long.';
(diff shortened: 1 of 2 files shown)


src/app/common/duck/utils.ts
```diff
@@ -47,6 +47,17 @@ export const handleSelfSignedCertError = (failedUrl: string, dispatch: any) => {
   dispatch(certErrorOccurred(failedUrl));
 };
 
+const testTargetNSName = (value: string) => {
+  if (value?.length < 3 || value?.length > 63) {
+    return 'The namespace name can be between 3 and 63 characters long.';
```
rayfordj, line 52:
```suggestion
    return 'The namespace name should be between 3 and 63 characters long.';
```
@ibolton336 , should this be **should** instead of **can** here?

rayfordj (changes requested):

src/app/common/duck/utils.ts
ibolton336: @rayfordj updated!

rayfordj: Thanks!

rayfordj (approved):
/lgtm

PR #1260, opened 2021-06-29
Reverse proxy

ibolton336:
Adds http-proxy-middleware dependency to handle a reverse proxy configuration for external requests to the cluster-api & discovery-api. These were the last pieces required to allow our MTC UI to run successfully on an IBM ROKS cluster. Previous work was done to allow custom proxy configuration via environment variables. This is an unrelated change and aims to resolve an outstanding cors issue that arises in the IBM environment due to the lack of ability to configure whitelisted routes within the IBM openshift cluster api server. 

With the reverse proxy in place, we can bypass cors all together by hitting (for example) 'cluster-api' from within the browser instead of a cross-origin domain. The node server will then pick up this request from the UI client and proxy it to the defined url specified in the config.dev.json file.


src/app/cluster/duck/sagas.ts
```diff
@@ -561,7 +565,11 @@ function* initDiscoveryCert() {
       yield delay(CLUSTER_RETRY_PERIOD);
       continue;
     }
-    const discoveryClient: IDiscoveryClient = ClientFactory.discovery(state);
+    const discoveryClient: IDiscoveryClient = DiscoveryFactory.discovery(
+      state.auth.user,
+      state.auth.migMeta.namespace,
```
seanforyou23, line 570: Not sure if it's related to this change, but I got the following warning when loading the app.

Screen Shot 2021-07-19 at 9 26 23 AM

Screen Shot 2021-07-19 at 9 26 10 AM

seanforyou23 (commented):
Changes look good to me. It's my first time loading up this app so I was mainly just looking to ensure nothing major looks broken.. I was able to load clusters, replication repositories, and plans without issue. The only thing I saw that may need attention is fetching namespaces. Do I need to do anything in particular to configure namespaces or is there an issue there?

src/app/cluster/duck/sagas.ts
seanforyou23: Ah, looks like I was missing a config for `discoveryApi` - updated and everything is working now!

seanforyou23 (approved):

ibolton336:
[mig-424](https://issues.redhat.com/browse/MIG-424)

PR #1269, opened 2021-07-21
Expose restore progress reporting in UI

shubham-pampattiwar:
This PR exposes the restore progress status in UI.


pranavgaikwad (changes requested):
What happens when Restore has PVR associated?

pranavgaikwad (approved):
Please ignore my previous comment, since these changes are for Restore Step which doesn't have any PVR associated, it only displays the consolidated progress bar

PR #1270, opened 2021-07-23
[MIG-562] Add version check for MTC operator version

ibolton336:
https://issues.redhat.com/browse/MIG-562?filter=-3

### DIFF
--- a/deploy/main.js
+++ b/deploy/main.js
@@ -40,6 +40,10 @@ let clusterApiProxyOptions = {
   },
   logLevel: process.env.DEBUG ? 'debug' : 'info',
   secure: false,
+  onProxyRes: (proxyRes, req, res) => {
+    proxyRes.headers['Content-Security-Policy'] = 'sandbox';
+    proxyRes.headers['X-Content-Security-Policy'] = 'sandbox';
+  },
 };
 
 let discoveryApiProxyOptions = {
@@ -50,17 +54,19 @@ let discoveryApiProxyOptions = {
   },
   logLevel: process.env.DEBUG ? 'debug' : 'info',
   secure: false,
+  onProxyRes: (proxyRes, req, res) => {
+    proxyRes.headers['Content-Security-Policy'] = 'sandbox';
+    proxyRes.headers['X-Content-Security-Policy'] = 'sandbox';
+  },
 };
 
 if (process.env['NODE_ENV'] === 'development') {
   clusterApiProxyOptions = {
     ...clusterApiProxyOptions,
-    secure: false,
   };
 
   discoveryApiProxyOptions = {
     ...discoveryApiProxyOptions,
-    secure: false,
   };
 }
 
--- a/src/app/common/duck/sagas.ts
+++ b/src/app/common/duck/sagas.ts
@@ -61,18 +61,36 @@ function* fetchMTCVersion(action: any): Generator<any, any, any> {
       packageManifestResource,
       'mtc-operator'
     );
-    const channels = packageManifestResourceResponse.data.status.channels.map(
-      (channel: any) => channel?.currentCSV
-    );
-    yield put(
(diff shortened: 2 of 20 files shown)


ibolton336:
image

vconzola (changes requested):
@ibolton336 This isn't what I had in mind when we talked about this in sprint planning yesterday. I'll provide a mockup, but probably not until next week. Also, did we find out if there's any risk of data loss by updating the operator? Whether there is or isn't, we'll want to tell the user. I'd also like to provide a link to a readme or something that tells the user what's new in the latest channel and high level instructions for updating. 

@eriknelson is there a single place we can direct the user to that shows what's new in each operator channel?

ibolton336:
Not a problem Vince. I expected there to be a few UX tweaks involved. Let me know what you have in mind if this is not it.

eriknelson:
@vconzola I don't know of any reliable link that allows us to point to the "latest changes".

@apinnick is there a reliable release notes URL that will have all of our release notes that we've been putting together for releases?

There is no data loss. You choose the next release channel (i.e., I'm on 1.4 and I edit the installed operator and choose 1.5). OLM will do an in-place upgrade of the operator and the operator will roll out the new operands. The CRDs are never removed, and neither are your resources like MigPlans.

Looking up the upgrade instructions, we are missing the instructions for changing channels (meaning upgrades from 1.4 -> 1.5). I think the instructions here only describe how to accept z stream updates like 1.4.5 -> 1.4.6. I will file a BZ to update (https://docs.openshift.com/container-platform/4.7/migrating_from_ocp_3_to_4/upgrading-3-4.html)

eriknelson:
@ibolton336 converted to draft PR to be clear this is still in progress.

apinnick:
> @apinnick is there a reliable release notes URL that will have all of our release notes that we've been putting together for releases?

@eriknelson  There is a reliable release notes URL but I just discovered that the Customer Portal version of MTC has not been synced for a month. I am looking into it and will let you know when the DPM fixes the problem.

vconzola:
@ibolton336 Here's a mockup of what the channel update alert should look like. Use a PF [inline alert](https://www.patternfly.org/v4/components/alert#inline-variations). Because it's a global alert, it should be placed in the page header, just below the title. The link to the release notes should open a new tab. Ideally, selecting the "Update operator" action will redirect to the OpenShift operator hub with MTC selected. Not sure if this is possible. If not, lemme know and I'll figure something else out for directing the user Operator Hub. 

@apinnick Please review the alert text and suggest changes. Thanks.

apinnick:
@vconzola I have reviewed the alert text. Do you need to mention "operator hub" if you have a link for updating the operator?

I am not sure how you can link to release notes. They are working on persistent URLS for Customer Portal docs but currently the URLs contain the version. So you can only link to MTC release notes by version, unfortunately. If the OpenShift version changes, the MTC link changes.

Suggested text: "MTC Operator v<version> is out of date. Upgrade to the latest version<'Update operator' link>. The upgrade does not affect stored data."

vconzola:
@ibolton336 Do you have everything you need for this now? Can we link directly to the MTC operator in operator hub from the alert as I describe above? If not, I want to add some text highlighting the update process.

ibolton336:
image

vconzola:
@apinnick We're using "update" and "upgrade" interchangeably here. I think we should use one or the other. Which is correct?

apinnick:
> @apinnick We're using "update" and "upgrade" interchangeably here. I think we should use one or the other. Which is correct?

@vconzola  

In this context, "upgrade" should be used because the user is changing versions (installing later version of Operator).

"Update" would not be the right word here because an update can include patching/fixing the current version, without necessarily upgrading it to a later version.

djwhatle (approved):
LGTM.

PR #1273, opened 2021-07-29
Add namespace to pod log command

ibolton336:
Fixes bug when namespace is not added to the oc logs command within the debug view.


pranavgaikwad (approved):
VISACK

PR #1274, opened 2021-07-29
Add missing reverse proxy forward

ibolton336:
For pod requests, we need to use the reverse proxy for discovery service as we are everywhere else


rayfordj (approved):
/lgtm

PR #1282, opened 2021-08-23
Bug xxxxxxx: select bug fix 

ibolton336:


ibolton336:
https://user-images.githubusercontent.com/11218376/130531929-9103bd67-57ce-4895-884b-f31ef9f8620f.mov

Previously, when all items in the table were selected, deselecting a single item would leave the select all checkbox still checked. This fixes this within the namespaces table and the PVs table

ibolton336:
@vconzola Does this look ok to you?

ibolton336:
Before fix:

https://user-images.githubusercontent.com/11218376/130532155-fc210028-41fc-405f-898a-151c027053ef.mov

vconzola (changes requested):
@ibolton336 You should use the PatternFly [bulk selector](https://www.patternfly.org/v4/guidelines/bulk-selection) component in the table header. This should handle checkbox state (and give a selection count) for you.

ibolton336:
> @ibolton336 You should use the PatternFly [bulk selector](https://www.patternfly.org/v4/guidelines/bulk-selection) component in the table toolbar. This should handle checkbox state (and give a selection count) for you.

@vconzola The patternfly component actually just gives us a ui component with a checkbox inside of a dropdown. We will still need to connect the component to our form state & handle logic for partial select/ select all. I propose that we open an enhancement for this that includes: 
- Removing select all functionality from composeable table 
- Adds the toolbar item for "bulk selection" that can be reusable across the app. This can be added to the FilterToolbar component that we currently use above the patternfly tables. 
- Tie in the suggested states outlined by the bulk selector to our internal form state

ibolton336:
https://github.com/konveyor/mig-ui/issues/1289

PR #1284, opened 2021-08-23
Bug 1996904: Maintain consistency among titles

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=1996904


ibolton336:
@vconzola How is this looking?

vconzola (changes requested):
@ibolton336 In order to get consistency in the different modals let's remove the plan name from the modal heading and just say the migration type followed by "migration". So,
- "Cutover migration"
- "Stage migration"
- "State migration"
- "Rollback migration"  with the type capitalized and the word "migration" lowercase.

eriknelson:
We have a ton of documentation, screenshots, labs, other content that rely on these options remaining named as they are today. I would prefer to delay changing these names until 1.7 when we have enough time to fully consider what those implications are.

apinnick:
> @apinnick are you able to help update the downstream documentation, wherever this may be required, and ensure that it's inline with the language we're using to describe these migrations in the UI?

@eriknelson 

"Cutover" is explained in the sections on [warm migration](https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.0/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#warm-migration_mtv), [creating a migration plan in the UI](https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.0/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#creating-migration-plan_mtv), [running a migration plan in the UI](https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.0/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#migrating-virtual-machines-cli_mtv), and [migrating VMs from the CLI](https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.0/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#migrating-virtual-machines-cli_mtv).

"Final migration" is not used in the MTV documentation.

However, the terms that are used in documentation are "precopy stage" and "cutover stage". "State migration" does not occur because the feature has not yet been documented. 

Are you saying that you want them changed to "Stage migration" and "Cutover migration"? I am not sure that makes things clearer because they are different stages of the same migration. In my opinion, calling it an "x migration" implies that it is a separate migration, not a separate stage or phase of the same migration. However, I can see how these terms would be useful shorthand in the UI. Personally, I think "precopy" is clearer than "stage", but if everyone prefers "stage", let me know and I will update the docs.

Interesting that "rollback" is not mentioned anywhere in the docs. When was the feature introduced?

eriknelson:
@apinnick I'm confused, you're referencing MTV? This is MTC? Until now, the literal actions on a migration plan where "Stage" and "Migrate". I'm concerned about calling it "Cutover" and confusing people. If you don't think that will be confusing to users and the documentation is squared away, then I'm satisfied. I just want to make sure everything is consistent between the product and the documentation.

apinnick:
> @apinnick I'm confused, you're referencing MTV? This is MTC? Until now, the literal actions on a migration plan where "Stage" and "Migrate". I'm concerned about calling it "Cutover" and confusing people. If you don't think that will be confusing to users and the documentation is squared away, then I'm satisfied. I just want to make sure everything is consistent between the product and the documentation.

@eriknelson @vconzola 

Oops. my bad. I was documenting RHV warm migration for MTV and got confused. 

You are correct. We use "Stage" and "Migrate" for MTC, not "Stage" and "Cutover".

image

vconzola:
@apinnick For MTC, as part of adding State only migrations in 1.6, I'd like to change the actions/kebab menu for migration plan to group the different migration types (Stage, State, Rollback, and Migrate) under a single "Migrations" heading. But Migrations - Migrate seems a more than a little weird so I suggested changing the term "Migrate" in the UI to "Cutover". This would make MTC consistent with MTV and is more descriptive of what the migration actually does. Marco has said he's OK with the change, and Ian has made the change in his UI PR. So we just need to update the docs. I do have some concern that the terminology change might be a little confusing to current MTC users, so it would be good to highlight it in release notes and the docs.

ibolton336:
Screen Shot 2021-08-25 at 12 21 13 PM
Screen Shot 2021-08-25 at 12 21 40 PM
Screen Shot 2021-08-25 at 12 22 05 PM
Screen Shot 2021-08-25 at 12 22 24 PM

PR #1286, opened 2021-08-24
Bug 1997665: Add paused polling effect and keys for state migration modal

ibolton336:
Fixes bug where selected items were changing due to background polling and resulting re-renders. Also adds keys to clean up debug dev console warnings. 

https://bugzilla.redhat.com/show_bug.cgi?id=1997665

### DIFF
--- a/src/app/home/pages/PlansPage/components/Wizard/WizardFormik.tsx
+++ b/src/app/home/pages/PlansPage/components/Wizard/WizardFormik.tsx
@@ -38,18 +38,12 @@ const WizardFormik: React.FunctionComponent<IWizardFormikProps> = ({
 
       if (!values.sourceCluster) {
         errors.sourceCluster = 'Required';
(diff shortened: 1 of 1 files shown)


eriknelson:
@ibolton336 I'm seeing a lot of new bugs crop up that seem like stuff you're seeing while moving about the dev UI? Do these all need to be cloned to 1.5 as well, or are they unique to 1.6?

Also, is this one just missing a BZ at all?

ibolton336:
> @ibolton336 I'm seeing a lot of new bugs crop up that seem like stuff you're seeing while moving about the dev UI? Do these all need to be cloned to 1.5 as well, or are they unique to 1.6?

This one is unique to 1.6 because it is dealing with the new state migrations modal. The others could be cloned to 1.5.z i think. 
> 
> Also, is this one just missing a BZ at all?

Yeah I haven't had the chance to create one for this yet.

PR #1298, opened 2021-09-01
Change 'can' to 'should' for consistency.

rayfordj:
@ibolton336 , This PR changes `can` to `should` to be consistent with existing messaging elsewhere.

Such as [src/app/common/duck/utils.ts#L52](https://github.com/konveyor/mig-ui/blob/master/src/app/common/duck/utils.ts#L52):
```
  if (value?.length < 3 || value?.length > 63) {
    return 'The namespace name should be between 3 and 63 characters long.';
  } else if (!DNS1123Validator.test(value)) {
```

Thanks!


ibolton336 (approved):
LGTM

PR #1324, opened 2021-09-08
Backport axios fix: 0.21.1 -> 0.21.4

rayfordj:


ibolton336 (approved):
Sort of strange that redux-saga types were added here. Shouldn't be an issue

rayfordj:
@ibolton336 , I realized I missed all the `start-server-and-test#wait-on` removal and would have still carried the old axios version.  I've updated & pushed which should now capture what we _**actually**_ want -- only axios-0.24.4.  Thanks!

PR #1337, opened 2021-09-17
Bug 2003603: Add dropdown for proper log selector pod name 

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=2003603


ibolton336:
image

image

vconzola:
@ibolton336 That drop down is kinda just floating out there and looks like it should be a list of pods. Can you put the pods into two columns and then put the dropdown below with a top aligned label, “Select cluster to view logs”?

PR #1340, opened 2021-09-21
Bug 2006316: Update node proxy configuration

ibolton336:
Opt for internal service routes instead of using the externally exposed discovery service api route.


jmontleon (dismissed):
ACK

PR #1351, opened 2021-09-29
Bug 2006316: Update node proxy configuration (#1340)

ibolton336:
* update proxy settings

* update url to pull from env

* Reinstantiate cluster API

* updates

* Add local dev support and revert discapi removal

* cleanup

* Add dev env var to keep development mode working

* update internal clusterSvcUrl


eriknelson:
@ibolton336 what's going on with this guy? seems like it's referencing the 1.6.0 bug and we might need to clone this back to 1.5.2?

PR #1355, opened 2021-10-27
Bug 2015550: Regression fix - respect NO_PROXY for token request

ibolton336:


eriknelson (approved):

eriknelson:
I verified the logic here with @shawn-hurley and worked with @ibolton336 to ensure that not only are domains and IPs covered if passed as the token endpoint, but we're also covered should a CIDR be listed in the NO_PROXY. We will check to see if the token endpint is actually an ip address whether or not it matches the range.

This should be solid and is passing all of our environment tests, thank you @ibolton336

PR #1361, opened 2021-11-12
Progressive wizard/storage class conversion implementation

ibolton336:
- Migration type support for wizard
- State migration refactor into wizard & away from standalone modal
- Storage class conversion support


src/app/plan/duck/selectors.ts
```diff
@@ -20,7 +20,7 @@ const getCurrentPlan = (state: DefaultRootState) => state.plan.currentPlan;
 
 const getCurrentPlanWithStatus = createSelector([getCurrentPlan], (currentPlan) => {
   if (currentPlan && currentPlan.status?.conditions) {
-    let statusObject = {};
+    let statusObject: any = {};
```
mturley, line 23: Note to self: can we use a better type than `any` here

mturley:
@eriknelson sorry for the delay in getting back to this, I've had a lot of interrupt-driven work this past week. I'm testing it now and found that if you try to edit an existing plan, the wizard crashes. Looking into it.

Screen Shot 2021-12-07 at 3 10 12 PM

mturley (approved):
Code LGTM for the create plan flow. Errors come up when editing a plan, and there are some differences from the mockups, so I opened https://github.com/konveyor/mig-ui/issues/1362 to follow up. Merging this so we can test the basic create flow end to end.

PR #1363, opened 2021-12-14
Address issues with progressive wizard / storage class conversion support

### LINKED ISSUE #1362: Follow up bugs after progressive wizard PR
Opening a separate issue to follow up after merging #1361 so we can start testing it end to end, since the main use case works.

Issues observed when testing the new wizard changes and comparing them to the [mockups](https://docs.google.com/presentation/d/1eO1tPsLQst7qbj-CKD10m4dzsxG3h7bRhAulqoKy40k/edit#slide=id.p):

- [x] Crashes when trying to edit existing plan (see https://github.com/konveyor/mig-ui/pull/1361#issuecomment-988232093)

- [x] General step
  - [x] "All fields are required" text is missing
  - [x] Fields are out of order, source cluster should be before migration type
  - [x] Too much space between the fields?
  - [x] "Repository" field should be titled "Replication repository"
- [x] Namespaces step
  - [x] In storage class conversion mode, there are editable row controls but nothing to edit
  - [x] Non-PF button styles in use for editable rows
- [x] PVs step
  - [x] migration type column should be disabled until a PV is selected (when selection is a thing, state+full only?)
  - [x] Target storage class field is missing from here, should be here instead of on copy options step when in storage class conversion mode (need to check with @vconzola)
  - [x] PV migration type is still here when in storage class conversion mode, it was removed in the mockups for that mode
- [x] Copy options step
  - [x] Non-PF styles in use for editable rows

Additional implementation needed:
- [x] Plans table
  - [x] Type column (Full, state only, storage class conversion) (Fixed in #1366)
  - [x] Rearrange table to fit everything? (compact variant? expandable w/ source + target under the fold?) (Fixed in #1366)
  - [x] Remove/rename actions in the dropdown as appropriate for each migration type
  - [x] Disable the Rollback action when it is not applicable (need to discuss with @eriknelson)

mturley:
Working towards #1362

### DIFF
--- a/src/app/home/pages/PlansPage/components/PlanActions/PlanActionsComponent.tsx
+++ b/src/app/home/pages/PlansPage/components/PlanActions/PlanActionsComponent.tsx
@@ -42,6 +42,7 @@ export const PlanActionsComponent: React.FunctionComponent<IPlanActionsProps> = 
     hasClosedCondition = null,
     hasReadyCondition = null,
     hasErrorCondition = null,
+    hasAttemptedMigration = null,
     hasRunningMigrations = null,
     finalMigrationComplete = null,
     isPlanLocked = null,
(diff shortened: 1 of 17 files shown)


mturley:
Ok, made some tweaks here but this is now ready to go. cc @eriknelson

pranavgaikwad (approved):
tested in my environment, overall looks good with minor issues that can be solved separately later.

PR #1364, opened 2021-12-16
Bug 2027382: Use timeouts on 'copied to clipboard' alerts, change alert timeout delay to PF-standard 8 seconds

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=2027382

PR #1365, opened 2021-12-16
Bug 2027382: [1.6.3 Backport] Use timeouts on 'copied to clipboard' alerts, change alert timeout delay to PF-standard 8 seconds

mturley:
Backports #1364


mturley:
cc @eriknelson

eriknelson (approved):

PR #1366, opened 2022-01-06
Add migration type column and filter on the plans table

mturley:
Adds a Type column to the plans table and adjusts widths and other layout transforms of the columns on the table to attempt to minimize wrapping. The state column is a little smaller now, so I made sure it will truncate properly with ellipses and a tooltip if it displays a longer message.

Also adds a Type filter to the filter dropdown and fixes a bug with FilterToolbar where select-mode filter dropdowns wouldn't open.

Screen Shot 2022-01-07 at 2 34 40 PM


mturley:
@vconzola since this one is just UI layout, would you mind approving?

vconzola (approved):
LGTM. Thanks @mturley.

PR #1367, opened 2022-01-06
Plan wizard: Prevent editing/validating plan name once the plan CR has been created

mturley:
Once the `currentPlan` is set in the redux store, the plan CR has been created and the wizard should lock in the plan name to prevent further changes or validation errors. You can observe this by proceeding to the PVs step of the wizard, then going back to the general step (the name field will be disabled).

https://issues.redhat.com/browse/MIG-874


mturley:
cc @eriknelson

JaydipGabani (approved):

PR #1371, opened 2022-01-12
Bug 2005998: Making sure targetName is accurate in namespace table wizard

JaydipGabani:


src/app/home/pages/PlansPage/components/Wizard/NamespacesTable.tsx
```diff
@@ -97,9 +97,25 @@ const NamespacesTable: React.FunctionComponent = () => {
     const editedNamespace = values.editedNamespaces.find(
       (editedNS) => editedNS.id === namespace.id
     );
+    const { currentPlan } = useSelector((state: DefaultRootState) => state.plan);
+    let isTargetNamespaceNameSet = false
+    let setTargetNamespaceName = ""
```
mturley, line 102: Can you use a non-verb name? Seeing references to `setTargetNamespaceName` may seem at first like it's a function. Maybe `targetNamespaceFromPlan` or something

src/app/home/pages/PlansPage/components/Wizard/NamespacesTable.tsx
```diff
@@ -97,9 +97,25 @@ const NamespacesTable: React.FunctionComponent = () => {
...
+    if (currentPlan != null) {
```
mturley, line 103: Try to avoid using `!=` or `==` in JS/TS, they perform type coercion that is usually undesirable. We should use `!==` or `===` instead.

In this case though, you can just do `if (currentPlan)` because a non-null plan is truthy.

src/app/home/pages/PlansPage/components/Wizard/NamespacesTable.tsx
```diff
@@ -97,9 +97,25 @@ const NamespacesTable: React.FunctionComponent = () => {
...
+      for (var i of currentPlan.spec.namespaces) {
+        let ns = i.split(":");
+        if (ns.length > 1) {
+          if (ns[0] == namespace.name) {
+            isTargetNamespaceNameSet = true
```
mturley, line 108: You probably don't need both of these variables, since `isTargetNamespaceNameSet` can be derived from whether or not `setTargetNamespaceName` is defined/truthy.

mturley (changes requested):
The logic here is sound, but I think it can be done a little cleaner. Something like this:

```ts
const { currentPlan } = useSelector((state: DefaultRootState) => state.plan);
let targetNamespaceNameFromPlan = null;
if (currentPlan) {
  currentPlan.spec.namespaces.forEach((planNamespace) => {
    const splitNs = planNamespace.split(':');
    if (splitNs.length > 1 && splitNs[0] === namespace.name) {
      targetNamespaceNameFromPlan = splitNs[1];
    }
  });
}

const isIntraClussterMigration = values.sourceCluster === values.targetCluster;
const initialTargetName = isIntraClusterMigration ? `${namespace.name}-new` : namespace.name;
const finalTargetName = targetNamespaceNameFromPlan || initialTargetName;
const targetNamespaceName = editedNamespace ? editedNamespace.newName : finalTargetName;
```

PR #1375, opened 2022-01-20
Properly show None selection for target storage classes, fix suggested target storage class init when creating a new plan

mturley:
Related to https://issues.redhat.com/browse/MIG-740

~~This reveals another issue it does not solve: that `|| pv.storageClass` default is misleading. When the user first sees this field (in Copy Options for full/state migrations, or in the PVs step for SCC) it will appear to default to the storage class of the source PV (even if that selection doesn't exist in the target), but really the form data is blank and a default will be used by the backend. The form used to default to a true selection determined by [this block of code](https://github.com/konveyor/mig-ui/pull/1361/files#diff-30964530bef8d26e28a5365204049ffb0fca889c73d890f805b0c0fa19c2f482L22-L54) that was removed in #1361. I'm not sure exactly why that was removed or what the implications would be of restoring it. It may be that instead we need to show this field as blank by default and require that the user make an explicit selection, or indicate that blank is a valid option and somehow convey what it will do. (note that this blank state is distinct from selecting None, which is a valid option)~~

This PR now also fixes the above issue by properly initializing these fields based on the selections in the Plan CR if they are not yet initialized upon PV discovery completion (i.e. when creating a new plan).

See slack discussion starting here for context: https://coreos.slack.com/archives/CHB7WSNUX/p1642699829044500

### DIFF
--- a/src/app/home/pages/PlansPage/components/Wizard/PVStorageClassSelect.tsx
+++ b/src/app/home/pages/PlansPage/components/Wizard/PVStorageClassSelect.tsx
@@ -48,9 +48,11 @@ export const PVStorageClassSelect: React.FunctionComponent<IPVStorageClassSelect
       onChange={(option: any) => onStorageClassChange(currentPV, option.value)}
       options={storageClassOptions}
       value={
-        storageClassOptions.find(
-          (option) => currentStorageClass && option.value === currentStorageClass.name
-        ) || pv.storageClass
+        currentStorageClass === ''
+          ? noneOption
+          : storageClassOptions.find(
+              (option) => currentStorageClass && option.value === currentStorageClass.name
+            ) || pv.storageClass
       }
       placeholderText="Select a storage class..."
     />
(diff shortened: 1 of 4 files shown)


sseago:
" it will appear to default to the storage class of the source PV (even if that selection doesn't exist in the target)," -- this seems wrong. That's not the way the back end logic works.

Back end logic for the initial value is as follows:
1) if it's gluster or NFS and ceph is available on dest, then the appropriate ceph storageclass is selected
2) else if a dest storageclass exists that matches the src storageclass provisioner, that will be selected
3) else the dest cluster's default storageclass will be selected (including "None" as an option if there is no default storageclass on dest)

That's all in the back end, though. From the UI point of view, you have a field with a current value and a set of valid values, so we need a dropdown with the valid values in the list, and the current value selected, just like any other dropdown that edits an existing object.The notion of "default" is less relevant from the UI point of view since the UI never *creates* anything here -- once we get to the storageclass selection field, we're editing fields rather than creating them.

mturley:
That makes sense. I think it's just a bug in the UI, it's not properly getting that existing selection when you create a new plan. "it will appear to default to the storage class of the source PV" is happening because the UI has a fallback to that value when it has nothing in form state, which it seems (a) shouldn't ever be possible and (b) isn't a sensible fallback anyway. I'll see if I can figure out what's going on with initializing properly from what's in the plan. Weirdly it seems to work correctly in edit mode, just not in create mode.

JaydipGabani (dismissed):
VISACK

mturley:
Rebased, can you re-approve @JaydipGabani ?
cc @eriknelson

JaydipGabani (approved):

PR #1387, opened 2022-02-15
Bug 2049576: making sure cancelled rollback behaves as epected

JaydipGabani:
Fixes [this](https://bugzilla.redhat.com/show_bug.cgi?id=2049576) bug

### DIFF
--- a/src/app/plan/duck/selectors.ts
+++ b/src/app/plan/duck/selectors.ts
@@ -144,7 +144,11 @@ const getPlansWithPlanStatus = createSelector(
       let planMigrations;
       if (plan.Migrations.length) {
         const latestRollbackIndex = plan.Migrations?.findIndex((m) => {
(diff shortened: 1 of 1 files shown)


eriknelson:
My bad, somehow that bz was missing a target release but I had it marked triaged. Updated!

src/app/plan/duck/selectors.ts
```diff
@@ -154,6 +158,8 @@ const getPlansWithPlanStatus = createSelector(
       } else {
         planMigrations = plan.Migrations;
       }
+      console.log('migration');
```
ibolton336, line 161: Can we remove these?

JaydipGabani: Done

mturley (approved):
LGTM, good catch

PR #1391, opened 2022-02-18
Bug 2055157: Fix pvs showing up empty after pv discovery

ibolton336:


src/client/resources/conversions.ts
```diff
@@ -472,12 +472,15 @@ export function updateMigPlanFromValues(
 
       return updatedPV;
     });
-    updatedSpec.persistentVolumes = planValues.persistentVolumes;
+    updatedSpec.persistentVolumes = _.union(
```
mturley, line 475: It looks like `_.union` will use referential equality to compare the objects, so if there is a PV that already appears in form state and is also in updatedSpec (a new copy of it, not the same reference), the PV would get included twice right? We might want to instead use [`_.unionBy()`](https://lodash.com/docs/4.17.15#unionBy) and give it a function for the second param that reduces each PV to a unique identifier

src/client/resources/conversions.ts
```diff
@@ -472,12 +472,15 @@ export function updateMigPlanFromValues(
 
       return updatedPV;
     });
-    updatedSpec.persistentVolumes = planValues.persistentVolumes;
+    updatedSpec.persistentVolumes = _.unionBy(
+      planValues.persistentVolumes,
+      updatedSpec.persistentVolumes
+    );
```
mturley, line 478: I think this needs to be:
```suggestion
    updatedSpec.persistentVolumes = _.unionBy(
      [planValues.persistentVolumes, updatedSpec.persistentVolumes],
      (pv) => pv.metadata.name
    );
```

src/client/resources/conversions.ts
ibolton336: Updated to include the iterator for the unique ID. @mturley

mturley (approved):

PR #1396, opened 2022-04-06
Bug 2072036: Allow storage class conversion plans to be created when no replication repositories are present

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=2072036

Currently the buttons to open the migration plan wizard are disabled if no replication repositories are found. However, storage class conversion plans do not require a replication repository.

This PR removes the logic to disable those buttons (the wizard can now always be opened) and instead disables the Full and State migration types in the General step of the wizard if no replication repositories are found.

The following tooltips appear when hovering over the disabled migration type options:

Screen Shot 2022-04-06 at 3 31 20 PM

Screen Shot 2022-04-06 at 3 31 25 PM


ibolton336 (approved):
Looking good!

PR #1399, opened 2022-04-07
Bug 2057516: Disable target PVC edit controls on Copy Options step for full migrations

mturley:
https://bugzilla.redhat.com/show_bug.cgi?id=2057516

Also corrects a related DOM structure issue where the number of `<Th>` cells in the table header could be different from the number of `<Td>` cells in each row.

Note: GitHub's diff is a little confused by the whitespace changes and thinks I made changes to the content of the `td.pf-c-table__inline-edit-action`, I only wrapped it in a conditional and indented it.


ibolton336 (approved):
LGTM thanks Mike

mturley:
Holding off merging this after a concern from Vince, will follow up.

PR #1436, opened 2022-06-03
Bug 2038898: Add valued update check for aws s3/s3 repo type

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=2038898


src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/S3Form.tsx
```diff
@@ -33,8 +34,44 @@ interface IOtherProps {
   initialStorageValues: any;
   provider: string;
   isAWS: boolean;
+  currentStorage?: any;
 }
 
+const valuesHaveUpdate = (values: any, currentStorage: IStorage, isAWS: boolean) => {
```
mturley, line 40: This is fine because it's the pattern we've established, but it has me wondering if we could use Formik better for this kind of thing. Isn't this what the `isDirty` property in form state is supposed to be for? Might be worth looking into why we didn't use that in the past (I suspect something to do with prefilling asynchronously, that's why I ended up adding a `prefill` function to useFormState)

mturley (approved):

PR #1444, opened 2022-06-14
Bug 2040693: Update repository name validation to use name length validation

ibolton336:


src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/AzureForm.tsx
```diff
@@ -35,26 +35,26 @@ const addEditButtonTextFn = addEditButtonText(componentTypeStr);
...
   if (!currentStorage) {
     return true;
-  }
-
-  const existingMigStorageName = currentStorage.MigStorage.metadata.name;
-  const existingAzureStorageAccount =
-    currentStorage.MigStorage.spec.backupStorageConfig.azureStorageAccount;
-  const existingAzureResourceGroup =
-    currentStorage.MigStorage.spec.backupStorageConfig.azureResourceGroup;
+  } else {
```
mturley, line 38: No need to hold this up, but these new `else` cases aren't needed because of the `return` statement in the conditions above. If it's more readable this way that's also fine.

mturley (approved):
LGTM, one nit that you can take or leave.

src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/AzureForm.tsx
ibolton336: @mturley I was paranoid here and added the else because I was seeing https://github.com/konveyor/mig-ui/pull/1444/files/e1471b0d9120b44a4b7f9c483c6ac81d794d9c10#diff-16f2c71744e2c1b667e3eb393f6ae76469593139ae762ed4d93a53b5736024d4R53

Cause a UI crash. The optional not was probably sufficient in this one place but just added the else to be sure that the app wont crash in the case of something under currentStorage being undefined. I can probably just remove the else statements and keep the option not operator on line 53.

PR #1447, opened 2022-06-14
Bug 2058172: Add plan status icon update for succeeded rollback

ibolton336:
https://bugzilla.redhat.com/show_bug.cgi?id=2058172

### DIFF
--- a/src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
+++ b/src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
@@ -23,6 +23,7 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
     hasNotReadyCondition = null,
     hasSucceededStage = null,
     hasSucceededCutover = null,
+    hasSucceededRollback = null,
     hasSucceededWithWarningsCondition = null,
     isPlanLocked = null,
     hasConflictCondition = null,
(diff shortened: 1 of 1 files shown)


src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
```diff
@@ -82,6 +84,12 @@ const PlanStatusIcon: React.FunctionComponent<IProps> = ({ plan }) => {
         <ResourcesAlmostEmptyIcon />
       </span>
     );
+  } else if (latestAction === 'rollback' && hasSucceededCondition) {
```
mturley, line 87: @ibolton336 you're gonna kill me, but I double checked and there is already a `hasSucceededRollback` on the `plan?.PlanStatus` that you can use instead of this line, just like `hasSucceededStage` and `hasSuccededCutover` above. See https://github.com/konveyor/mig-ui/blob/master/src/app/plan/duck/selectors.ts#L189-L198

I think that's actually more accurate than this because `hasSucceededCondition` is based on conditions on the plan and not the migration (I'm not sure if the plan gets a success condition when a rollback succeeds but the rollback's migmigration definitely does)

mturley (changes requested):

src/app/home/pages/PlansPage/components/PlanStatusIcon.tsx
ibolton336: Oops, I see that hasSucceededRollback checks that the latest action is a rollback.

mturley: yeah it's not the best naming :)

mturley (approved):
LGTM!

PR #1454, opened 2022-06-16
Bug 2038898: Adjust s3 initial values to match expected

ibolton336:


src/app/home/pages/StoragesPage/components/AddEditStorageModal/ProviderForms/S3Form.tsx
```diff
@@ -55,9 +55,9 @@ const valuesHaveUpdate = (values: any, currentStorage: IStorage, isAWS: boolean)
     }
 
     const existingS3URL = currentStorage.MigStorage.spec.backupStorageConfig.awsS3Url;
-    const existingRequireSSLValue = currentStorage.MigStorage.spec.backupStorageConfig.insecure;
+    const existingRequireSSLValue = !currentStorage.MigStorage.spec.backupStorageConfig.insecure;
```
ibolton336, line 58: The requireSSL value needs to be !insecure as it is here:
https://github.com/konveyor/mig-ui/blob/186cc4f97cc178096b18348b33282ac02a1a24e0/src/app/home/pages/StoragesPage/components/AddEditStorageModal/AddEditStorageForm.tsx#L53 
when the initial value is set. 

This is causing the isDisabled check to always show false since these 2 values are not matching up in the case of an s3 repo. None of the other repos allow manual setting of the "insecure" value, so that is why this bug only shows up for s3 repos. 

Same story for the customCA value. If the value does not exist, hard set it to null instead of allowing it to be undefined for comparison purposes.

mturley (approved):

PR #1469, opened 2022-08-08
Move devDependencies over to main deps

ibolton336:
Fixes the build error resulting from using yarn --production flag in docker file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     "@typescript-eslint/eslint-plugin": "^4.18.0",
     "@typescript-eslint/parser": "^4.18.0",
     "clean-webpack-plugin": "^4.0.0",
-    "compression": "^1.7.4",
     "concurrently": "^5.2.0",
     "css-loader": "^5.0.1",
     "eslint": "^7.22.0",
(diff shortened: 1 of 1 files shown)


rayfordj (approved):
/lgtm

mturley:
Hmm, @ibolton336 should we review the rest of the devDependencies to see if anything else needs to be moved over? I can imagine something like that sneaking up on us at run time

ibolton336:
> Hmm, @ibolton336 should we review the rest of the devDependencies to see if anything else needs to be moved over? I can imagine something like that sneaking up on us at run time Edit: gave them a once-over and nothing else jumps out at me

@mturley I did a quick check and built images to test. Looked like those were the only two that were causing the image to fail to build. You're instinct to check further gaps is spot on though. Not sure if some other functionality would maybe lazy load during run time like you mentioned. I just looked through the main.js entry point file.

mturley:
I looked at each of the dependencies left in that list and looked up the ones I didn't recognize, they're all of the build/lint/test/mock variety. I think we're probably good.

PR #1471, opened 2022-08-24
Replace `@konveyor/lib-ui` with `@migtools/lib-ui`

mturley:
The `@konveyor/lib-ui` package has been migrated from the `konveyor` org to the `migtools` org as part of Konveyor being [donated to the CNCF Sandbox](https://www.konveyor.io/blog/konveyor-is-a-cncf-sandbox-project/). The npm package has been renamed to `@migtools/lib-ui` starting with the `8.4.1` release (see https://github.com/migtools/lib-ui/pull/111).

Upgrading lib-ui from 5.1.2 -> 8.4.1 involves the following breaking changes:

* [From 6.0.0](https://github.com/migtools/lib-ui/releases/tag/v6.0.0):
  * **useformstate:** Forms that rely on `!field.isValid` to show errors will now show validation errors on all required fields as soon as the form first renders. Consumers should replace the `!field.isValid` condition with `field.shouldShowError` or `!field.isValid && field.isTouched`.
* [From 7.0.0](https://github.com/migtools/lib-ui/releases/tag/v7.0.0):
  * **ValidatedTextInput:** Any direct calls to getTextInputProps or getFormGroupProps that were passing a boolean for greenWhenValid will now need to instead pass an options object including greenWhenValid as a property.
* [From 8.0.0](https://github.com/migtools/lib-ui/releases/tag/v8.0.0):
  * **useFormState:** Any forms using the yupOptions argument will need to wrap it in an object with the property yupOptions instead of passing it directly as a top-level argument.

None of these changes affect this repo because it is not using `useFormState` or `ValidatedTextInput`, so renaming imports is the only source change necessary here.

See also:
* https://github.com/konveyor/forklift-ui/pull/984
* https://github.com/konveyor/forklift-console-plugin/pull/27
* https://github.com/migtools/crane-ui-plugin/pull/134
* https://github.com/konveyor/tackle2-ui/pull/375

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,10 +74,8 @@
     "webpack-dev-server": "^4.8.1"
   },
   "dependencies": {
-    "https-proxy-agent": "^5.0.0",
-    "compression": "^1.7.4",
     "@hot-loader/react-dom": "^17.0.1",
-    "@konveyor/lib-ui": "^5.1.2",
+    "@migtools/lib-ui": "^8.4.1",
     "@patternfly/react-core": "^4.115.5",
     "@patternfly/react-icons": "^4.9.5",
     "@patternfly/react-styles": "^4.8.5",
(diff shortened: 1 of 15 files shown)


mturley:
Edit: actually, realizing the other change in this upgrade was a bump in the PatternFly versions used as peer dependencies in lib-ui. Marking this as a draft, may not return to it considering the status of mig-ui.

PR #1514, opened 2024-07-22
Bug 2269583: Resolve follow-redirects to fixed version

ibolton336:
- Axios minor version bump to include fix https://github.com/axios/axios/pull/6163 
- Resolve resolve follow-redirects to fixed version 1.15.6


rayfordj (approved):
/lgtm

PR #1522, opened 2024-08-15
remove package-lock.json

awels:
Add package-lock.json to gitignore so it doesn't
accidetnally get added to the repo.


.gitignore
```diff
@@ -7,9 +7,10 @@ yarn-error.log
 config/config.dev.json
 .yalc/
 yalc.lock
+package-lock.json
```
mateusoliveira43, line 10: usually, its bad idea to ignore this file, right? 

would add comment about the error that having this file causes

rayfordj: @mateusoliveira43 , 

downstream MTC UI is failing scratch builds
```
Error in plugin resolve_remote_source: Cachito request is in "failed" state, reason: File check failed for yarn: the package-lock.json file must not be present
```

```
Fetching the source from "https://github.com/migtools/mig-ui" at reference "51e0ff92cb62209fdf8e7558f8334f9234231a37"
...
cachito.errors.InvalidRepoStructure: File check failed for yarn: the package-lock.json file must not be present
```

[PR#1521](https://github.com/migtools/mig-ui/pull/1521) appears to introduce a net new package-lock.json file.  Looking at its history, the last time it was present was 2019.

rayfordj (approved):
/lgtm

No complaints from Cachito & a successful scratch build with the change.

```
[2024-08-15 14:24:47,501 celery.app.trace INFO trace.info] Task cachito.workers.tasks.general.finalize_request[13692c35-2b8e-4b50-aebf-0652aac22685] succeeded in 2.8828462900128216s: None
```

```
Completed:	Thu, 15 Aug 2024 10:35:51 EDT
```

.gitignore
awels: So we want to gitignore this file so it doesn't get checked into the repository. It will get generated when building.

PR #1526, opened 2024-10-17
Behave the same as 1.8.4 on a new migration plan

awels:
The loading behavior was modified in 1.8.5 and
the grid loading was not behaving the same as
1.8.4 when creating a new direct volume migration
plan. The spinner was not covering the grid and
the grid was showing the empty state.

Instead revert to covering the entire grid. And
only show the validating spinner when the grid
selection changed.

Show warning in plan grid if live migration is selected, but kubevirt version is not new enough or has not enabled storage live migration


dymurray (approved):
VISIACK

